PR #5, opened 2017-08-17
Small Correction. 

ralle12345:
addRule cannot be imported from rules, but automationManager can.


steve-bate:
You should pull the latest updates. The addRule import is correct. The openhab.rules module doesn’t export automationManager any more because the module must look it up dynamically from the script-specific scope when addRule is invoked. I changed it after I noticed that rules were not being deleted when a script was reloaded. The addRule function should be equivalent to:

PR #11, opened 2018-01-13
Switch to Log4j logging and using logger for output exceptions

allan-gam:
I've done various changes which involves switching from slf4j logging to Log4j. I've also made changes to how the error that gets caught in the exception clauses are put out. It will help a lot when using the Karaf console to monitor the log files in real time. 

The root log name has been changed from "" to "jython.scripting". Probably better to use the value from the constant `LOG_PREFIX = "org.eclipse.smarthome.automation.rules"` However it's so long so it will get truncated in the log.

Please review the suggested changes carefully if you consider to merge them. I'm a newbie and I make newbie mistakes. Thanks.

### DIFF
--- a/lib/openhab/log.py
+++ b/lib/openhab/log.py
@@ -4,7 +4,7 @@ import traceback
 
 from org.apache.logging.log4j import Logger, LogManager
 
-LOG_PREFIX = "org.eclipse.smarthome.automation.rules"
+LOG_PREFIX = "sboh2j"
 
 def log_traceback(fn):
     """Decorator to provide better Jython stack traces"""
@@ -16,7 +16,7 @@ def log_traceback(fn):
             if len(args) > 0 and hasattr(args[0], "log"):
                 args[0].log.error(traceback.format_exc())
             else:
-                logger = LogManager.getLogger("jython.scripting")
+                logger = LogManager.getLogger(LOG_PREFIX)
                 logger.error(traceback.format_exc())
             raise # Re-raise the exception (allowing a caller to see and handle the exception as well)
     return wrapper
@@ -26,7 +26,7 @@ class Log4j2Handler(logging.Handler):
         message = self.format(record)
         logger_name = record.name
         if record.name == "root":
-            logger = LogManager.getLogger("jython.scripting")
+            logger = LogManager.getLogger(LOG_PREFIX)
         else:
             logger = LogManager.getLogger(logger_name)
         level = record.levelno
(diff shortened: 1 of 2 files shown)


steve-bate:
Thanks for the PR, but can you provide more context about the changes? Has openHAB2 stopped using the slf4j adapter for log4j? What currently happens if the Python logging.handler uses Logger.ROOT_LOGGER_NAME for the Logger root name?

allan-gam:
Sure. 
According to Rich Koshak "openHAB 2 uses a completely different logging library and approach to logging than openHAB 1.x does, replacing logback with log4j" Have a look [here](url). Even though log4j seems to work in some aspects I believe it's a good thing to use the log4j, especially if wanting to being able to configure log levels etc.

I also believe that **sboh2j** (steve-bate/openhab2-jython) should have its own root namespace, not being a child to "org.eclipse.smarthome.automation.rules". By having that it would allow users to define both a logger and appender for **sboh2j**. I also suggest adding a short note about how to add the **sboh2j** entry to the `userdata/etc/org.ops4j.pax.logging.cfg`file.

I make an additional change in my repo branch. It's just a suggestion. You may of course reject any part or edit it in any way you prefer.

steve-bate:
The link you included doesn't work. Doing a search on the Smarthome source code, they still use slf4j extensively so I'm interested in seeing the post you mentioned. I searched the forums and haven't seen any other reference to the change in November (but the forum search is horrible so I may have missed it). Note that slf4j is not a logging framework itself. It's a wrapper for other logging frameworks. Using slf4j supports easy switching from logging frameworks like log4j to java.util.logging, for example. The logback framework is just another implementation that directly uses the slf4j API. Maybe you saw an announcement about switching from logback to log4j? If so, that doesn't mean OH2 stopped using slf4j.

Why does the "org.eclipse.smarthome.automation.rules" prefix prevent defining loggers and appenders? Or are you saying it would make it easier?

steve-bate:
I think I found the post. Rich is literally saying they replaced logback with log4j. That's not the same as saying they replaced slf4j with log4j (as I described earlier). 

https://community.openhab.org/t/taming-openhab-2-logging/13976

steve-bate:
@allan-gam, I've added logging to the function decorators although the approach was a little different than what you proposed. Thanks for reporting the print statement that had been left there previously. I'm still using the automation.rules log category prefix, but I add the function name to make it easier to diagnose error log messages and will support function-specific logging level configuration. It is also now possible to specify a custom logging suffix in the decorator using the `trigger_name` optional keyword argument.

PR #15, opened 2018-07-14
Module with class that simulates executeCommandLine with a timeout

openhab-5iver:
I'm note sure if it is best to add this as a separate module, but it didn't seem to fit into an existing one. I'm also not sure if this is a good name for it. Usage is...
```
from openhab.command import Command
rawMessage = Command(['/usr/bin/curl','-s','-X','GET','https://api.wunderground.com/api/<WU_API_key>/alerts/hourly/conditions/q/pws:KOHBRUNS25.json']).run(timeout=10)[0]
```
Comparable to this in the Rules DSL...
```
    val String rawMessage = executeCommandLine("/bin/sh@@-c@@/usr/bin/curl -s -X GET https://api.wunderground.com/api/<WU_API_key>/alerts/hourly/conditions/q/pws:KOHBRUNS25.json",10000)
```

### DIFF
--- a/lib/openhab/command.py
+++ b/lib/openhab/command.py
@@ -1,0 +1,21 @@
+# This modue simulates executeCommandLine with a timeout. It will return a tuple (stdoutdata, stdouterr).
+import subprocess, threading
+
+class Command(object):
+    def __init__(self, cmd):
+        self.cmd = cmd
+        self.process = None
+        self.data = None
+
+    def run(self, timeout):
+        def target():
+            self.process = subprocess.Popen(self.cmd,stdout=subprocess.PIPE)
+            self.data = self.process.communicate()
+
+        thread = threading.Thread(target=target)
+        thread.start()
+        thread.join(timeout)
+        if thread.is_alive():
+            self.process.terminate()
+            thread.join()
+        return self.data


besynnerlig:
Thanks for this lovely contribution.

If I understand correct, the benefit over just using `os.system(cmd)` is that you can retrieve the command result. Mostly when I do system calls, I call curl with different parameters like the lucid example below. 

Sometimes a cascade of curl commands can arise and I don't want to block all my event threads (it would stall openHAB) so I typically send all my jobs to background without checking the results. Ideally there would be a callback URL that triggered a special type of event that we could do some scripting for. (Like the "Asynchronous HTTP Request and handling"  they have in [dzVents](http://www.domoticz.com/wiki/DzVents:_next_generation_LUA_scripting#logging_.3D_.7B_..._.7D_.28optional.29) for Domoticz)

I believe that's a brilliant solution. 

Anyway, do you think your executeCommandLine should have an option to not wait for the result but to send the job to bg?

```python
import os
import lucid.config as config

def autoremote(message, ttl=300, sender='openHAB'):
    '''
    Sends an autoremote message
    '''

    # Use GCM Server for delivery
    cmd = 'curl -s -G "https://autoremotejoaomgcd.appspot.com/sendmessage" ' \
        + '--data-urlencode "key='+config.autoremote['key']+'" ' \
        + '--data-urlencode "password='+config.autoremote['password']+'" ' \
        + '--data-urlencode "message='+message+'" ' \
        + '--data-urlencode "sender='+sender+'" ' \
        + '--data-urlencode "ttl='+str(ttl)+'" ' \
        + ' 1>/dev/null 2>&1 &'

    os.system(cmd)
```

openhab-5iver:
> If I understand correct, the benefit over just using os.system(cmd) is that you can retrieve the command result.
I'm not familiar with os.system(cmd)... I'm still a noob with Jython/Python. My intent was to provide the same functionality that is available in the Rules DSL using executeCommandLine.

> and I don't want to block all my event threads 
Is this a concern with JSR223? It was my understanding (not sure how/where it came from :smiley:) that JSR223 would not be limited by the thread pools. Similar to bindings. I haven't tested this or looked into the code though. I have noticed that a rule instance is blocking, and will prevent other instances of the rule from starting until it has finished.

> Ideally there would be a callback URL that triggered a special type of event that we could do some scripting for.
I see this as ideal... I'll look into it. I'll also look into whether os.system(cmd) could just call a Python script, which includes an API call to update/command an item.

> Anyway, do you think your executeCommandLine should have an option to not wait for the result but to send the job to bg?
Sounds like a good 2.0! Or another function for a utils.py. I've been also researching [multiprocessing](https://docs.python.org/2/library/multiprocessing.html), but I'm not clear on how OH would react to it.

STOP THE PRESSES!
I keep forgetting I'm in Jython! By adding this import...

    from org.eclipse.smarthome.model.script.actions.Exec import executeCommandLine

I can use executeCommandLine directly. So this PR is unnecessary!

PR #16, opened 2018-07-14
Combine item_group_triggered and item_triggered into the same decorator

openhab-5iver:
This change combines the two decorators, and adds the allMembers and groupAsItem arguments so that this single decorator...

1. Accepts a single item as the item_name (non-group item default)
2. Accepts a group item, using .getAllMembers (with allMembers=True)
3. Accepts a group item, using .getMembers (with allMembers=False)
4. Accepts a group item as an item (groupAsItem=True)

It made sense to me to combine all of this into one, rather than split them out into multiple decorators. Without any arguments other than item_name, item_triggered will create a single item trigger for items or individual item triggers for all members of a group. Use groupAsItem=True to create a rule that triggers on the state of a group changing. Use allMembers=False to create a rule with triggers for only a groups direct children (similar to 'Member of' in the Rules DSL).

### DIFF
--- a/lib/openhab/triggers.py
+++ b/lib/openhab/triggers.py
@@ -140,11 +140,12 @@ def time_triggered(cron_expression, trigger_name=None):
     return decorator
 
 ITEM_CHANGE = "ItemStateChangedEvent"
+GROUP_CHANGE = "GroupItemStateChangedEvent"
 ITEM_UPDATE = "ItemStateEvent"
 ITEM_COMMAND = "ItemCommandEvent"
 
-def item_triggered(item_name, event_types=None, result_item_name=None, trigger_name=None):
-    event_types = event_types or [ITEM_CHANGE]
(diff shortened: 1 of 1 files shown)


openhab-5iver:
> I've looked at your PR and i think you should put statements on line 165 to multiple lines for readability?

@ricadelic, respectfully,  I think introducing a new variable and spreading it to multiple lines makes it less readable.
```
for i in (item.getAllMembers() if allMembers else item.getMembers()):
```
vs.
```
groupMembers = []
if allMembers:
    groupMembers = item.getAllMembers()
else:
    groupmembers = item.getMembers()
for i in groupMembers:
```
I prefer the ternary, but as long as the functionality is there, I'm good with it either way! :smiley:

openhab-5iver:
I have a much larger change to triggers.py that will supersede this one, so I'm removing this PR.

PR #27, opened 2018-10-21
Minor documentation edits - Please Review

sillyfrog:
A couple of minor edits of things I found and have tried to fix. Not 100% sure it's the best way, so if you can review that would be great.

Also, It would be great if you were able to turn on the Wiki on the repo, that way people could put up examples etc is a much more casual way (or if there is a better place, maybe put up with Wiki read-only and link to it there).

Thanks again for your work on this, it's great - Python is soooo much better than Xbase! :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,22 @@ In order to use them, these modules will need to be copied to a subdirectory of 
 
 </ul>
 
+## Custom Modules
+<ul>
+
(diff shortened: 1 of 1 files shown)


openhab-5iver:
Thank you!

I thought about a Wiki, but figured the forum would be a better place. I've requested a category and subcategories be added for JSR223, Jython, JavaScript, Groovy, etc., but it's been taking a loooong time. Although, it may be easier for people to contribute in a Wiki. Commits can be done from the web, so that sounds easy too. Please add a discussion issue and see if the other maintainers will share an opinion. If not, we can flip a coin...

PR #31, opened 2018-10-30
Add logging instance to execute function

gitMiguel:
This adds logging argument into the execute function which is created from rules name. This way every rule has automatically its own logger even if they are defined in the same file:
```python
from openhab.rules import rule
from openhab.triggers import when

@rule("TestRule")
@when("Item Test changed")

def execute(event, log): 
    trigger = event.itemName
    log.info("Triggered by: \"{}\"".format(trigger))


@rule("TestRule2")
@when("Item Test changed")

def execute2(event, log):
    trigger = event.itemName
    log.info("Triggered by: \"{}\"".format(trigger))
```
Produces:

```python
2018-10-30 16:42:43.755 [INFO ] [ipse.smarthome.model.script.TestRule] - Triggered by: "Test"
2018-10-30 16:42:43.757 [INFO ] [ipse.smarthome.model.script.TestRule2] - Triggered by: "Test"
```

Like to hear your thoughts if this is something that suits "core" features.

Signed-off-by Miika Jukka <miikajukka@gmail.com>


openhab-5iver:
Hello, Mike. I'm sorry for such a delay, but I was very busy over Halloween, and then my whole family came down with a very nasty cold/flu that we are still fighting off.

Did you close the PR due to issues with the implementation? I was planning to look into the openhab.log module, which I do not use and am not familiar with the functionality it provides. I wasn't sure if there may be some crossover between it and what you had added.

gitMiguel:
I closed this because I didn't get any feedback and merging changes from master didn't go so smooth. I made a mistake working from two different local repos.

You take your time and go through openhab.log. I'm not familiar with it either. These proposed changes was just to provide a easier start when people not so familiar with jython/python starts to write their rules.

gitMiguel:
Forgot to say that I'm willing to continue this if you like. But then I'll create a clean branch with latest changes merged and maybe a separate discussion.

PR #32, opened 2018-10-30
Corrected mispelling of 'Descendent' in triggers.py

openhab-5iver:


shutterfreak:
The proper spelling is with an 'a' as in `descendants`.
I would suggest to deprecate the use of `descendents` and to introduce `descendants`.

PR #43, opened 2018-11-30
Dockerfile

diijkstra:
I'm very thankful for your work, these scripts are both good examples how to convert rules to Python, as well as make life easier. Since getting jython to run takes quite a lot of steps, I've decided to create a Dockerfile which will install Jython with all needed environment variables onto official Docker image of Openhab. 

Image was tested on debian version of amd64 and armhf flavours with M4, M5 and M6 images. It requires recent addition to official images which is a directory with scripts run on each startup. On each startup, it will enable experimental rule engine addon in addons.cfg and append python options to EXTRA_JAVA_OPTS.

openhab2-jython is not installed via dockerfile because, probably everybody will want to have their version in container's volumes for easier editing. One can easily write docker-compose on top of this Dockerfile.

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -1,0 +1,23 @@
+## From which base container as in https://hub.docker.com/r/openhab/openhab/#image-variants eg. 2.4.0.M6-armhf-debian
+ARG OPENHAB_VERSION
+FROM openhab/openhab:${OPENHAB_VERSION}
+
+## Where will we install jython inside container
+ARG JYTHON_HOME="/opt/jython"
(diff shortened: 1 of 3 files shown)


diijkstra:
Nah, don't worry about delay! I've rebased my branch against master and capitalized docker directory. I would even consider to moving it to Community, what do you think?

About standalone jar, to be honest, I've forgot about that option. On the other hand, with full install, one can use pip inside container (once issue around 2.7.1 will be resolved).

openhab-5iver:
> I would even consider to moving it to Community, what do you think?

I think it is good where it is.

PR #55, opened 2018-12-15
Updated core/items.py to use ItemBuilderFactory service

ScottKinSF:
I have created this PR to update core/items.py to use the newItemBuilder() method of the ItemBuilderFactory service rather than the deprecated itemRegistry.newItemBuilder() method.

I have tested it on my installation and it is working without the deprecation warnings.


openhab-5iver:
LGTM

PR #61, opened 2018-12-29
class to make using item metadata easier

mjcumming:
Proposed class to make using item metadata easier. Class is designed to work with a single metadata namespace. Provides methods to read/write the Value for the namespace and the Configuration (key/value) pairs for the namespace.

Thoughts?

### DIFF
--- a/Core/automation/lib/python/community/metadata_item_namespace.py
+++ b/Core/automation/lib/python/community/metadata_item_namespace.py
@@ -1,0 +1,73 @@
+'''
+
+Class to simplify reading and writing item metadata
+
+Instantiate with item name and the metadata namespace to work with
+
+'''
+
+from org.eclipse.smarthome.core.items import Metadata
+from org.eclipse.smarthome.core.items import MetadataKey
+from core import osgi
+
+MetadataRegistry = osgi.get_service("org.eclipse.smarthome.core.items.MetadataRegistry")
+
+class Metadata_Item_Namespace(object):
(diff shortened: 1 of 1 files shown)


smhgit:
@mjcumming 

Have you thought of adding namespace manipulation functions to the class, such as:

```
set_namespace - change item namespace context
namespaces - list of all item namespaces
delete_namespace - delete existing namespace
```

This also means that constructor can have only item name

mjcumming:
Yes, was thinking about how to best approach that.

In my use case, I am always interacting with an single namespace of an item. It keeps the complexity of working with metadata lower. Having to set the namespace context could be problematic. The methods could be modified to always require a namespace.

The existing class does allow for deletion (remove) of the namespace.

I can’t think of any use case where one would want to enumerate all of the namespaces of an item?

smhgit:
@mjcumming 

I agree that there is no immediate use for namespace functionality, so it can be postponed until real need.

Just to clarify, you don't need a name space in each function, the set_namespace function was for that purpose, to switch namespace, the functions should continue to work as they are.

mjcumming:
@openhab-5iver, this is ready to merge unless there are any concerns.

smhgit:
Any plans to merge it?

openhab-5iver:
@mjcumming, are you OK with me closing this, as it looks like everything is in #88?

mjcumming:
It's been too long for me to remember the difference between then. My occupy code is this so it works for it's purpose .

PR #63, opened 2019-01-11
Refactor items.py and provider, and add links.py and provider, and small fix to log.py

openhab-5iver:
This will introduce a breaking change for items.py. There is some namespace conflict with the default scope, but I think we can work around that. But the ItemChannelLinkProvider is needed for removing Items, since the links need to be removed before the Item can be removed. When I also added links.py, it was getting messing having core.items.add and core.links.add, so I renamed them to core.items.add_item and core.links.add_link (same for remove). Take a look at the Community OWM script to see some links.py usage. This PR is needed to get it working..

I was also getting exception when removing Items that hadn't gone through the JythonItemProvider, and I think I got that working now. 

There were some missing {} in log.py, preventing exceptions from being logged properly. This will hopefully get the exceptions from rules logged properly. In a previous commit, I had added a CloudNotification on exceptions, and I saw these coming in now with this change.

### DIFF
--- a/Core/automation/jsr223/core/components/200_JythonItemChannelLinkProvider.py
+++ b/Core/automation/jsr223/core/components/200_JythonItemChannelLinkProvider.py
@@ -1,0 +1,45 @@
+from org.eclipse.smarthome.core.thing.link import ItemChannelLinkProvider
+
+import core
+import core.osgi
+
+class JythonItemChannelLinkProvider(ItemChannelLinkProvider):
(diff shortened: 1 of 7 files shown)


ScottKinSF:
Commit 23abc95 LGTM, but I would like to understand the change of `withTags(tags)` to `withTags(set(tags))` in items.py. Did that change come about from reading through the related ESH source?

PR #69, opened 2019-01-12
Fix exception that occurred when tags parameter is not supplied to add_items()

ScottKinSF:
Changed default value of tags from `None` (not iterable) to `list()` (iterable).


ScottKinSF:
Sure thing. I honestly don't know which way of specifying an empty list is preferred in the python world, I'm more of a C++ guy.

openhab-5iver:
[] is faster! :smile:  https://stackoverflow.com/a/2972226

PR #73, opened 2019-01-19
Ide alarm migration

fastlorenzo:
Updated with latest master changes + added migrated utils from lucid needed for ideAlarm to work


openhab-5iver:
I've included your changes in #91, so this can be closed.

PR #85, opened 2019-03-02
cancelled-Documentation for date upgrades

CrazyIvan359:
And here is some documentation for the changes I made to the `core.date` module.

I had a bit of trouble getting my fork to line up with your master branch I think. Sorry if it's a bit messy, this has been my first foray into using git.

### DIFF
--- a/Community/OpenWeatherMap/automation/jsr223/community/openweathermap/owm_daily_forecast.py
+++ b/Community/OpenWeatherMap/automation/jsr223/community/openweathermap/owm_daily_forecast.py
@@ -8,27 +8,32 @@ the Items and groups. As the day progresses, there will be fewer hourly
 forecasts included in the gForecast_1 group.
 
 You may want to comment out several groups and Items that you do not plan on
-using, but the Timestamp Items are required for the script to function. The
+using, but the Forecast_Timestamp Items are required for the script to function. The
 rule will also run when the script file is saved, or OH is restarted, so that
-you don't have to wait for the trigger.
+you don't have to wait for the trigger for the Item states to be populated.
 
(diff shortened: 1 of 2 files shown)


openhab-5iver:
The doc looks good, but you'll need to cleanup your branch to get rid of the other commits. I think all you need is to rebase. If you're not comfortable doing this in git, you can edit the file and submit the PR directly in Github.

CrazyIvan359:
Great.

I will learn. I should probably have learned git a long time ago. Did some digging and seems I should have fetched your repo and pushed that to my fork, then made the changes in a new branch. I will try that when I get home. Thanks for your patience.

I'm also working on a metadata module (started from the one in the other pull request). I will get docs done and everything before I do a pull for that though

openhab-5iver:
> Thanks for your patience.

No problem!

> I'm also working on a metadata module (started from the one in the other pull request).

I really _really_ need to get that one finalized... it's long overdue. I saw some issues with it though, and I'm totally caught up in other projects. If you are looking to make changes to it, please just submit them to the branch for that PR and discuss with @mjcumming. IIRC, I have a whole page written up , but OTTOMH, there is no documentation, and a provider should be used (which I think I wrote already). Anyhow, if you get some sparks flying over there, I'll get back into it! :smile: I'm also about to finalize [this one](https://github.com/openhab/openhab-core/pull/519).

CrazyIvan359:
I was hesitant to add to that branch since I have almost completely rewritten it. The classes are self managed now and behave essentially like a Python dict. I will do a commit with what I have so far. No docs but everything has descriptions in the functions.

As for the openhab-core stuff, can I help? I may be much good with git yet, but I think I'm an ok programmer. I'm certainly on board with the idea of having Jython et al be built in

PR #88, opened 2019-03-06
Add Metadata classes

CrazyIvan359:
As I mentioned, metadata classes!

@openhab-5iver I don't have write access on #61 so I couldn't push to that branch. I figured since mine differs so much from that one I would start my own branch and PR for it.

@mjcumming Credit where credit is due Michael, I started writing this module with your code in #61 as a reference point and it was a great help as I had not done anything with metadata before this. I found your class to be rather raw in that you always manually read and write to the registry. I moved to a more hands off approach where the class immediately loads when created and writes when any changes are made. My `metadata_namespace` class loosely resembles your class, and I added another simple class to allow managing multiple namespaces of a single item from one object.

&nbsp;
> and a provider should be used (which I think I wrote already).

I forgot you had mentioned this until I was looking at the other PR to compose this message. I don't see anything metadata in the codebase. 
Presently I am not using a Provider, but I found the code in OH Core. I will look into this more, though at present I'm not entirely sure it's necessary for the purposes of this class. As I understand it, the Provider allows OH to know what attributes a namespace should have, but that seems beyond the scope of this module as it stands.
I could create a Provider that would allow adding namespace definitions from JSR223 code, this would be useful for what I am getting into as it would allow changes to the namespaces from PaperUI.

### DIFF
--- a/Core/automation/lib/python/core/metadata.py
+++ b/Core/automation/lib/python/core/metadata.py
@@ -93,7 +93,7 @@ class metadata_namespace(MutableMapping):
     def __len__(self): return len(self._configuration)
     def __setitem__(self, key, value): return self.set_config_value(key, value)
     def __getitem__(self, key): return self.get_config_value(key)
-    def __delitem__(self, key): self.delete_config_value(key)
+    def __delitem__(self, key): return self.delete_config_value(key)
 
     def load(self):
         '''Loads the namespace from the metadata registry.
@@ -122,12 +122,12 @@ class metadata_namespace(MutableMapping):
         # save to the registry
         MetadataRegistry.add(Metadata(self._keyUID, str(self._value), strConfiguration))
         del strConfiguration
-    
+
     def remove(self):
         '''Deletes this namespace from the registry.
         This instance will be preserved and can be written back to the registry.'''
         MetadataRegistry.remove(self._keyUID)
-    
+
     def set_value(self, value, save=True):
         '''Sets the namespace "value".
         Set "save" flag to False to skip saving to openHAB. You must take care to manually save
@@ -135,9 +135,19 @@ class metadata_namespace(MutableMapping):
         self._value = value
         if save: self.save()
 
+    @value.setter
+    def value(self, value):
+        '''Sets the namespace "value" and saves to the registry.'''
+        return self.set_value(value)
+
     def get_value(self):
         '''Gets the namespace "value".'''
         return self._value
+
+    @property
+    def value(self):
+        '''Returns the namespace "value".'''
+        return self.get_value
 
     def set_config_value(self, key, value, save=True):
         '''Sets the namespace configuration value for the specified key.
--- a/Core/automation/lib/python/core/metadata.py
+++ b/Core/automation/lib/python/core/metadata.py
@@ -8,11 +8,13 @@ from org.eclipse.smarthome.core.items import Metadata, MetadataKey
 from core.log import logging, LOG_PREFIX
 log = logging.getLogger(LOG_PREFIX + ".core.metadata")
 from core import osgi
-MetadataRegistry = osgi.get_service("org.eclipse.smarthome.core.items.MetadataRegistry")
+metadata_registry = osgi.get_service("org.eclipse.smarthome.core.items.MetadataRegistry")
 
 __all__ = [ "namespace_exists", "item_metadata", "metadata_namespace" ]
 
-# valid characters for names
+# Valid characters for names
+# this allows only numbers and letters for first char, 
+# then numbers, letters, and underscore for the rest
 _valid_chars_re = "^[a-zA-Z][a-zA-Z0-9_]*$"
 
 
@@ -22,7 +24,7 @@ def namespace_exists(item_name, namespace):
     '''
     if not re.match(_valid_chars_re, namespace):
         raise ValueError("'{}' is not a valid namespace name".format(namespace))
-    return False if MetadataRegistry.get(MetadataKey(namespace, item_name)) is None else True
+    return False if metadata_registry.get(MetadataKey(namespace, item_name)) is None else True
 
 def _item_exists(item_name):
     '''
@@ -99,7 +101,7 @@ class item_metadata(MutableMapping):
         if name in self._namespaces:
             raise KeyError
         else:
-            self._namespaces[name] = metadata_namespace(self._item_name ,name)
+            self._namespaces[name] = metadata_namespace(self._item_name, name)
 
     def delete_namespace(self, name, remove=False):
         '''
@@ -112,7 +114,6 @@ class item_metadata(MutableMapping):
             if remove:
                 self._namespaces[name].remove
             self._namespaces.pop(name, None)
-
 
 class metadata_namespace(MutableMapping):
     def __init__(self, item_name, name, value=None, configuration={}, load=True, save=False):
@@ -146,7 +147,7 @@ class metadata_namespace(MutableMapping):
             log.debug("Metadata: Save on initialize for item '{item}' namespace '{namespace}' with value '{value}' and configuration'{configuration}'".format( \
                 item=item_name, namespace=name, value=str(value), configuration=str(configuration)))
             self.save()
-    
+
     def __iter__(self):
         return iter(self._configuration)
 
@@ -170,7 +171,7 @@ class metadata_namespace(MutableMapping):
         log.debug("Metadata: Getting metadata for item '{item}' from namespace '{namespace}'".format( \
             item=self._item_name, namespace=self._name))
         # read from registry
-        metadata = MetadataRegistry.get(self._key_uid)
+        metadata = metadata_registry.get(self._key_uid)
         log.debug("Metadata: {metadata}".format( \
             metadata=str(metadata)))
         # parse data
@@ -188,15 +189,12 @@ class metadata_namespace(MutableMapping):
         unless the "save" flag was set to false when making those changes)
         '''
         # convert all values to strings
-        if self._value is None:
-            strValue = "None"
-        else:
-            strValue = str(self._value)
+        strValue = str(self._value)
         strConfiguration = {}
         for key, value in self._configuration:
             strConfiguration[key] = str(value)
         # save to the registry
-        MetadataRegistry.add(Metadata(self._key_uid, strValue, strConfiguration))
+        metadata_registry.add(Metadata(self._key_uid, strValue, strConfiguration))
         del strConfiguration
 
     def remove(self):
@@ -204,7 +202,20 @@ class metadata_namespace(MutableMapping):
         Deletes this namespace from the registry.
         This instance will be preserved and can be written back to the registry.
         '''
-        MetadataRegistry.remove(self._key_uid)
+        metadata_registry.remove(self._key_uid)
+
+    def get_value(self):
+        '''
+        Gets the namespace "value".
+        '''
+        return self._value
+
+    @property
+    def value(self):
+        '''
+        Returns the namespace "value".
+        '''
+        return self.get_value
 
     def set_value(self, value, save=True):
         '''
@@ -222,19 +233,6 @@ class metadata_namespace(MutableMapping):
         Sets the namespace "value" and saves to the registry.
         '''
         return self.set_value(value)
-
-    def get_value(self):
-        '''
-        Gets the namespace "value".
-        '''
-        return self._value
-
-    @property
-    def value(self):
-        '''
-        Returns the namespace "value".
-        '''
-        return self.get_value
 
     def set_config_value(self, key, value, save=True):
         '''
--- a/Core/automation/lib/python/core/metadata.py
+++ b/Core/automation/lib/python/core/metadata.py
@@ -3,83 +3,141 @@ Wrapper class for using Item Metadata
 '''
 
 from collections import MutableMapping
+import re
 from org.eclipse.smarthome.core.items import Metadata, MetadataKey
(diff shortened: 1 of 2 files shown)


CrazyIvan359:
> Presently I am not using a Provider, but I found the code in OH Core. I will look into this more, though at present I'm not entirely sure it's necessary for the purposes of this class. As I understand it, the Provider allows OH to know what attributes a namespace should have, but that seems beyond the scope of this module as it stands.
I could create a Provider that would allow adding namespace definitions from JSR223 code, this would be useful for what I am getting into as it would allow changes to the namespaces from PaperUI.

So I dug around in openhab-core and understand Providers/Registries a little more now. They do not provide templates, but I'm still not sure they are the right approach for this class.
If I use a custom provider we would only have access to metadata that had been created in Jython code. The `MetadataRegistry` seems to implement a built in `ManagedMetadataProvider` that looks after any add/update/remove requests and saves to the jsondb. While this approach makes sense for most use cases when thinking of items and thing creation in scripts, I don't think its the best approach for metadata.

mjcumming:
@CrazyIvan359 changes to metadata look great.

openhab-5iver:
Once lucid is merged, lets get this one merged and then go after the documentation.

mjcumming:
Let me know what is needed

CrazyIvan359:
@openhab-5iver I think this is ok to merge, let me know if you need me to make any changes though

Core/automation/lib/python/core/metadata.py
```diff
@@ -0,0 +1,193 @@
...
+        If "load" is set the namespace will be loaded from the registry. If it 
+        exists "value" and "configuration" will be overwritten if provided. 
+        If "save" is set the namespace will be written immediately to the 
+        registry. This will overwrite the namespace if it exists already. 
+        If the "load" flag is set the "save" flag will be ignored.'''
+        log.debug("Metadata: Initializing namespace object for item '{item}' namespace '{namespace}'".format( \
+            item=item_name, namespace=name))
+        self._item_name = item_name
+        self._name = name
+        self._keyUID = MetadataKey(self._name, self._item_name)
```
openhab-5iver: _key_uid

Core/automation/lib/python/core/metadata.py
```diff
@@ -0,0 +1,193 @@
...
+
+__all__ = [ "namespace_exists", "item_metadata", "metadata_namespace" ]
+
+
+def namespace_exists(item_name, namespace):
+    '''Returns True if metadata namespace exists'''
+    return False if MetadataRegistry.get(MetadataKey(namespace, item_name)) is None else True
+
+def _resolveType(value):
+    '''Attempts to resolve the type of the metadata value.
```
openhab-5iver: My preference is to start and end block comments on their own lines too. I think it is easier to read and identify as a comment. WDYT? Although, if we [stuck to the standard](https://www.python.org/dev/peps/pep-0008/#block-comments), we'd be using #s.

Core/automation/lib/python/core/metadata.py
```diff
@@ -0,0 +1,193 @@
...
+        else: self._namespaces[name] = metadata_namespace(self._item_name ,name)
+
+    def delete_namespace(self, name, remove=False):
+        '''Deletes the specified namespace from this object. It will not
+        remove the namespace from openHAB unless the "remove" flag is set'''
+        if name not in self._namespaces: raise KeyError
+        else:
+            if remove: self._namespaces[name].remove
+            self._namespaces.pop(name, None)
+
```
openhab-5iver: extra empty line

Core/automation/lib/python/core/metadata.py
```diff
@@ -0,0 +1,193 @@
...
+        return True
+    elif str(value).lower().strip() == "false":
+        return False
+    else:
+        # attempt to cast to int
+        try: return int(str(value))
+        except ValueError: pass
+        # attempt to cast to float
+        try: return float(str(value))
+        except ValueError: pass
```
openhab-5iver: Sorry if this is nit picky, but I think these should go on separate lines for readability. There are some other places too.

Docs/Jython-Modules.md
```diff
@@ -197,6 +197,43 @@ core.link.remove_link("Kodi_Control")
 ```
 </ul>
 
+#### Module: [`core.metadata`](..Core/automation/lib/python/core/metadata.py)
+<ul>
+
+This module allows simpler interaction with item metadata. There are two instance classes that can be used manage a single namespace, or multiple namespaces of a single item.
```
openhab-5iver: In an md, put sentences on separate lines with a space after the period.

Core/automation/lib/python/core/metadata.py
CrazyIvan359: Oops right beside some lower_case ones too lol. I think I did this while testing and reflexively used camel case since it's my go to

Core/automation/lib/python/core/metadata.py
CrazyIvan359: Like this?
```python
def _resolve_type(value):
    '''
    Attempts to resolve the type of the metadata value.
    '''
````

I have no preference really. I tend to have them on the same line to save on line count. I use VSC so with the highlighting my eyes just ignore quotes when they see green.

File: Core/automation/lib/python/core/metadata.py (hunk 3 above)
Comment (CrazyIvan359) [reply Core/automation/lib/python/core/metadata.py]: Like this for readability?
```python
        else:
            if remove: self._namespaces[name].remove
            
            self._namespaces.pop(name, None)
```

Core/automation/lib/python/core/metadata.py
CrazyIvan359: Consistency is important. This is my style, not the repo's. Thinking about it, I don't know where I picked up the habit of trying to keep the line count down

Docs/Jython-Modules.md
CrazyIvan359: I'll make the changes, but purely out of curiosity, why?

CrazyIvan359:
> This looks well written. Have you fully tested it though? I had to do a PR recently for core.date, due to a number of errors we came across in the lucid migration.

I have not fully tested it yet, no. I also forgot that I never did tests of all of `core.date` either. I don't have a working dev setup for openHAB + Jython right now though, thanks to the ESH merge, so testing is a bit difficult for me. Do you know of a way to `apt` install _any_ build of OH? I can only see about a dozen recent builds and don't think I found an archive of build packages anywhere. Even better would be your branch that fixes all of the import names post ESH merge. I haven't had much time to work on a computer in the last couple weeks, though I might have enough time to test tonight.

When I wrote this, I did use it to read and correctly type all types it recognises. It reads value and the config array correctly. I only used the namespace class, not the item class in those tests.

> Some simple examples at the top of the Jython-Modules.md entry would be helpful... like adding, reading, removing metadata from an Item.

Good idea. I'll clean it up.
  
  
I will finish up the page I'm writing in the docs and get to work on this

Core/automation/lib/python/core/metadata.py
```diff
@@ -0,0 +1,193 @@
+'''
+Wrapper class for using Item Metadata
+'''
+
+from collections import MutableMapping
+from org.eclipse.smarthome.core.items import Metadata, MetadataKey
+from core.log import logging, LOG_PREFIX
+log = logging.getLogger(LOG_PREFIX + ".metadata")
+from core import osgi
+MetadataRegistry = osgi.get_service("org.eclipse.smarthome.core.items.MetadataRegistry")
```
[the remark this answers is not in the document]
CrazyIvan359: Should we not keep this name as is since it isn't ours? I used this case because its a name from OH

CrazyIvan359:
I've made some of the changes, waiting on your input on others. Also cleaned up a few things, added some sanity checks on namespace names, checking for item existing, and added `None` to recognised types.

I did not have time to setup a test env to do further tests yet and may not have time to work on this tomorrow, so I've pushed the changes here in case you wanted to work on it.

Core/automation/lib/python/core/metadata.py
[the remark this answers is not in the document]
openhab-5iver: Actually, I just found that the precedent has already been set in core.actions to use snake_case. So, something like metadata_registry.

Core/automation/lib/python/core/metadata.py
openhab-5iver: > Like this?

Yes, that is what I was meaning. Visually, it's more 'blocky', which IMO is easier to read. I use VSC too, but it was tough to read it in GH!

Core/automation/lib/python/core/metadata.py
openhab-5iver: No, no. :slightly_smiling_face: There are two empty lines there, so I was suggesting to remove one.

Core/automation/lib/python/core/metadata.py
openhab-5iver: I tend to do the same. In the past, I would try to condense everything as much as I could, but I'm becoming more used to the concept of readability.

Docs/Jython-Modules.md
openhab-5iver: If there's one big blob of text on a line, it's harder to track changes. openhab-docs requires per sentence tracking too. I think there might even be a SAT check for this.

Core/automation/lib/python/core/metadata.py
CrazyIvan359: Oh I see, I wondered about that while I was working on it last night. I usually do that between classes/functions and single line spaces between methods or subfuncs so it's easier to see the separation. I was inconsistent with that in the file though. So a preference for single blank lines throughout then?

openhab-5iver: Yes, but I'm not sure if PEP8 says anything about it.

Core/automation/lib/python/core/metadata.py
```diff
@@ -0,0 +1,297 @@
...
+        self._value = value
+        if save:
+            self.save()
+
+    @value.setter
+    def value(self, value):
+        '''
+        Sets the namespace "value" and saves to the registry.
+        '''
+        return self.set_value(value)
```
openhab-5iver: Testing the first example (metadata_namespace), I get an error when using the module. Removing this function fixed it, but I did not test all methods yet. There is another function named `value` at L233, but I think this issue was with `value.setter`. The error was...

```
2019-05-12 10:59:48.624 [ERROR] [org.openhab.core.automation.module.script.internal.ScriptEngineManagerImpl] - Error during evaluation of script 'file:/opt/openhab2/conf/automation/jsr223/python/personal/test/testScript.py': NameError: name 'value' is not defined in <script> at line number 53
```

CrazyIvan359: I think this is one I added after I had done my tests. I think I know what it's complaining about though. When defining custom properties as methods like that you have to start with the getter then have the setter. If memory serves (I'm on my phone) it's like this:

```python
@property
def value(self):
    return self._value

@value.setter
def value(self, value):
    self._value = value
```

I may have forgotten to add that getter. I'll look at it tonight

CrazyIvan359: Oops, you said there is another at line 233, that's the issue, the getter needs to be first

Core/automation/lib/python/core/metadata.py
CrazyIvan359: That's fair, I hadn't really noticed on GH

Core/automation/lib/python/core/metadata.py
CrazyIvan359: PEP8 says what I do, but should we stick to one blank for consistency?

> Surround top-level function and class definitions with two blank lines.
>
> Method definitions inside a class are surrounded by a single blank line.
>
> Extra blank lines may be used (sparingly) to separate groups of related functions. Blank lines may be omitted between a bunch of related one-liners (e.g. a set of dummy implementations).
>
> Use blank lines in functions, sparingly, to indicate logical sections.

Core/automation/lib/python/core/metadata.py
```diff
@@ -188,23 +189,33 @@ def save(self):
         unless the "save" flag was set to false when making those changes)
         '''
         # convert all values to strings
-        if self._value is None:
-            strValue = "None"
-        else:
-            strValue = str(self._value)
+        strValue = str(self._value)
```
CrazyIvan359: Turns out `str(None)` gets you `'None'` so I removed this extra check

CrazyIvan359:
@openhab-5iver was this just waiting on testing?

openhab-5iver:
Documentation, mainly examples, and testing. It was difficult for me to use it. I also question why we'd want the caching, which complicates things. This one is next, after the doc rewrite.

mjcumming:
> Documentation, mainly examples, and testing. It was difficult for me to use it. I also question why we'd want the caching, which complicates things. This one is next, after the doc rewrite.

I have kept with the implementation I did a pull request for - its not as versatile but its simpler and fits better with my needs.

CrazyIvan359:
Well the caching is for simplicity I suppose, reading from the registry is a bit involved. Also consider the life-cycle of the class, it will only be around for the time it takes a rule to run.

mjcumming:
> Well the caching is for simplicity I suppose, reading from the registry is a bit involved. Also consider the life-cycle of the class, it will only be around for the time it takes a rule to run.

I depends on your use case - I have scripts that constantly refer to metadata  and change it as required. Caching would introduce some potential issues that would be hard to track down.

CrazyIvan359:
This module has undergone major changes and a new PR has been made (#167)

PR #92, opened 2019-04-28
Add .gitignore and name change "But-How-Do-I...-%3F.md"

besynnerlig:
1) .gitignore helps prevent uploading of files that should remain local.

2) Renaming the documentation file "But-How-Do-I...-%3F.md" to a file name that's valid also on a Windows file system.


.gitignore
```diff
@@ -0,0 +1,6 @@
+*.py[cod]
+*$py.class
+configuration.py
+!configuration.py.example
+personal/
```
openhab-5iver: Core/automation/jsr223/personal/**
Core/automation/lib/python/personal/**

We don't want to ignore the directory, just the files inside it. And it's theoretically possible that there could be another personal directory somewhere, so we could be specific about the path.

.gitignore
```diff
@@ -0,0 +1,6 @@
...
+jython-standalone-2.7.0.jar
```
openhab-5iver: What if we add the /automation/jython/ directory to the repo, and just ignore its contents?
```
Core/automation/jython/**
```

Docs/README.md
```diff
@@ -19,4 +19,4 @@ These works are based on the original contributions of Steve Bate, for which we
     - [Using Jython Extensions](Defining-Rules.md#using-jython-extensions)
         - [Rule and Trigger Decorators](Defining-Rules.md#rule-and-trigger-decorators)
     - [Event object attributes](Event-Object-Attributes.md)
-- [But how do I...?](But-How-Do-I...-%3F.md)
\ No newline at end of file
+- [But how do I...?](But-How-Do-I.md)
```
openhab-5iver: It would be best to have this in a separate PR, to keep it separate in the history. Not a big deal, but...

.gitignore
besynnerlig: Good point!

.gitignore
besynnerlig: Way better! ;)

Docs/README.md
besynnerlig: I don't have a working environment for working with git and I've messed things up. You can ignore this and we can add it in a separate PR later when I get a grip of the situation. I totally agree that it's not related. I comitted it earlier to my repo and I didn't know how to unstage/uncommit it.

openhab-5iver: Working in VS Code, there's an Undo Last Commit. In Eclipse, there's an Amend Commit option. From command line, go to the git directory and...

git commit --amend

But that only works for the last commit. You'd need to rebase since there have been multiple commits. Don't worry about it.

openhab-5iver:
I think we'll need to add a README.md file to the /automation/jython/ directory for it to show up. I had to do that for some other empty directories.

openhab-5iver:
@besynnerlig, you did not sign your commits. I added this to the repo recently, as this would be something required if/when this becomes an official OH repo. I'll merge this in without, but keep this in mind for the future.

PR #99, opened 2019-05-14
[utils] The full item object is logged instead of the item's name

besynnerlig:
See: https://github.com/OH-Jython-Scripters/openhab2-jython/issues/98

### DIFF
--- a/Core/automation/lib/python/core/utils.py
+++ b/Core/automation/lib/python/core/utils.py
@@ -112,10 +112,10 @@ def postUpdateCheckFirst(itemName, newValue, sendACommand=False, floatPrecision=
         if item.state != compareValue or (type(newValue) is float and floatPrecision is not None and round(item.state.floatValue(), floatPrecision) != newValue):
             if sendACommand:
                 sendCommand(item, newValue)
-                log.debug("New sendCommand value for [{}] is [{}]".format(itemName, newValue))
+                log.debug("New sendCommand value for [{}] is [{}]".format(item.name, newValue))
             else:
                 postUpdate(item, newValue)
-                log.debug("New postUpdate value for [{}] is [{}]".format(itemName, newValue))
+                log.debug("New postUpdate value for [{}] is [{}]".format(item.name, newValue))
             return True
         else:
             return False


openhab-5iver:
Thank you! What do you think of the refactoring I did to postUpdateCheckFirst?

besynnerlig:
> Thank you! What do you think of the refactoring I did to postUpdateCheckFirst?

Very neat! I've copied it into my production environment for testing it extensively. I appreciate your work!

besynnerlig:
@openhab-5iver 

Wherever possible I'll try to prepare my old scripts before migration from lucid to the new openhab2-jython and OH 2.5M1.

There's no document summarizing the changes that breaks my scripts is there? I believe not. Maybe you can recall some steps that you had to go through when migrating the lucid scripts? If you do, would you like to hinting me about the most important things? There's probably a few more people that will try to do the same migration.

openhab-5iver:
1) The `openhab` package was renamed to `core`.
2) core.rules.rule is now a callable (that returns a decorator), so in it's most minimal form, you need to use it like `@rule()` (note the parenthesis). Previously, it was not a callable (`@rule`).
3) core.triggers no longer registers the rule. core.rules.rule does that.
4) When defining triggers in classes (this was the big breaking change in the API), append `.trigger`. Check the [`hello_world.py`](https://github.com/OH-Jython-Scripters/openhab2-jython/blob/master/Script%20Examples/hello_world.py) for examples.
5) To catch the logging, you may need to set a logging level for 'jsr223.jython'. I setup an appender for it, so that it goes to a separate file than openhab.log. I have another for 'jsr223.jython.core' too. I plan to update the docs about logging. I also plan to change LOG_PREFIX to just 'jython' in the PR for 2.5.

That's all that come to mind right now. If you are migrating your rules to use the decorators, 1 and 5 are the only ones you'd need to be concerned with.

besynnerlig:
> 1. The `openhab` package was renamed to `core`.
> 2. core.rules.rule is now a callable (that returns a decorator), so in it's most minimal form, you need to use it like `@rule()` (note the parenthesis). Previously, it was not a callable (`@rule`).
> 3. core.triggers no longer registers the rule. core.rules.rule does that.
> 4. When defining triggers in classes (this was the big breaking change in the API), append `.trigger`. Check the [`hello_world.py`](https://github.com/OH-Jython-Scripters/openhab2-jython/blob/master/Script%20Examples/hello_world.py) for examples.
> 5. To catch the logging, you may need to set a logging level for 'jsr223.jython'. I setup an appender for it, so that it goes to a separate file than openhab.log. I have another for 'jsr223.jython.core' too. I plan to update the docs about logging. I also plan to change LOG_PREFIX to just 'jython' in the PR for 2.5.
> 
> That's all that come to mind right now. If you are migrating your rules to use the decorators, 1 and 5 are the only ones you'd need to be concerned with.

Thanks a million! That will really help me get started.

PR #137, opened 2019-05-27
[javascript] Fix typo in README.md

HerzScheisse:


HerzScheisse:
Maybe we should also mention in the Readme that by default the logger level is set to WARN...
It took me quite a while to realize why my logInfo wasnt printing anything to the log:tail 

```
log:set INFO jsr223.javascript
```

openhab-5iver:
Thank you! Please open an issue for the logging level and another to remove the reference to JSRule in the readme (I'm on a phone).

PR #142, opened 2019-05-31
Update utils.js with getTriggeringItemStr

HerzScheisse:
Dont know if its used to be in this core utils.js file or should went to personal folder...

I added this function to my utils.js file because its very usefull for me...

If you think it should not be here... feel free to delete this PR :)

source of this code snippet:
https://github.com/illnesse/openhab2_pub_001/blob/master/automation/jsr223/00_jslib/helper.js

### DIFF
--- a/Core/automation/lib/javascript/core/utils.js
+++ b/Core/automation/lib/javascript/core/utils.js
@@ -301,7 +301,14 @@
 			}
 		}
 	};
-	
+
+	context.getTriggeringItemStr = function(input)
+	{
+		var ev = input.get("event")+"";
+		var evArr = ev.split("'").join("").split("Item ").join("").split(" ");
+		return evArr[0];
+	};
+
 	//### getTriggeredData ###
 	context.getTriggeredData = function(input) {
 		


openhab-5iver:
Wouldn't you just use event.itemName? If so, then I do not think this would be useful in utils.js. But I could be wrong! 

I have a documentation update PR that I hope to get out today, where I've pulled together all of the common information shared by each language. Like this doc, which is pertinent here... https://github.com/openhab-scripters/openhab-helper-libraries/blob/master/Docs/Python/Event-Object-Attributes.md#event-object-attributes.

HerzScheisse:
Thank you for your comment. yeah i just need to get the triggered item name to split the last diget number of that item.
i do it now with:
`var triggeringItem = getItem(getTriggeringItemStr(input));`

but yeah you are right... i can achieve this without my modifications in this PR

`var triggeringItem = input.event.itemName;`

I'm just started working with the new rule engine... 

thank you :)

openhab-5iver:
Excellent! My understanding then is that this PR is not needed, so closing. If you have any other questions, feel free to @ me in the forum or just tag your post with jsr223.

PR #143, opened 2019-06-04
Fix string triggers and UUIDs

### LINKED ISSUE #141: [triggers] Error creating trigger UID when new_state contains quotes
**Describe the bug**

I've run into an issue whilst trying to convert some of my DSL rules into jython rules.  

I am using the following rule decorator which when parsed causes an invalid rule:

`@when("Item SONYReceiverSTRDN1080_Zone1_InputSource changed to \"netService:audio?service=spotify\"")`

The error given is:

`(INVALID_RULE): Validation of rule 144040b7-9cc8-4db4-95b0-6ad7fbadd717 has failed! Invalid module uid: Item-SONYReceiverSTRDN1080_Zone1_InputSource-changed-to-"netService:audio?service=spotify". It is null or not fit to the pattern: [A-Za-z0-9_-]*`

It appears that the library is trying to convert the when clause into a trigger uid.  As my when clause contains escaped quotes (which is valid) when converted into a trigger uid it violates the given regex pattern ([A-Za-z0-9_-]*) governing valid trigger uids.

Have looked through the code briefly (and not being a python expert) I think the issue is on this line or the following lines:

https://github.com/openhab-scripters/openhab-helper-libraries/blob/master/Core/automation/lib/python/core/triggers.py#L236

**To Reproduce**

Create the following jython rule and place in the personal directory:

```
from core.rules import rule
from core.triggers import when

@rule("Change Sony AVR Sound Field (Spotify)", description="This is an example Item changed triggered rule using decorators", tags=["Exa$
@when("Item SONYReceiverSTRDN1080_Zone1_InputSource changed to \"netService:audio?service=spotify\"")
def hellowWorldItemUpdateDecorators(event):
    hellowWorldItemUpdateDecorators.log.info(event.itemName + " hello world" )
```

**Expected behavior**
The library creates a rule with a valid trigger uid.

**Environment (please complete the following information):**
 - Jython version: [e.g. 2.7.0 standalone]
 - openHAB version: [e.g. 2.5.0 S1603]

openhab-5iver:
Triggers for StringItems that used a new_state or old_state could have invalid characters for the Trigger UUIDs, which was raising an exception in the validation. This PR corrects that, and also allows for states and Channel triggers with spaces in them. While working this PR, I could see extreme edge cases where the Trigger UUIDs were not unique and the Triggers would overwrite themselves. To prevent this, all Trigger names now have generated UUIDs appended to them. 

You can specify a StringItem's state with the following formats. This is only necessary if the state or Channel trigger has spaces in it.

@when("Item Virtual_String_1 changed from 'this is a test' to 'this is another test'")
@when("Item Virtual_String_1 changed from \"this is a test\" to \"this is another test\"")
@when('Item Virtual_String_1 changed from "this is a test" to "this is another test"')
@when('Item Virtual_String_1 changed from \'this is a test\' to \'this is another test\'')

Fixes #141 

### DIFF
--- a/Core/automation/lib/python/core/triggers.py
+++ b/Core/automation/lib/python/core/triggers.py
@@ -4,6 +4,8 @@ scope.scriptExtension.importPreset(None)
 import inspect
 import json
 import uuid
+import re
+from shlex import split
 
 import java.util
 from java.nio.file.StandardWatchEventKinds import ENTRY_CREATE, ENTRY_DELETE, ENTRY_MODIFY
@@ -42,7 +44,10 @@ log = logging.getLogger(LOG_PREFIX + ".core.triggers")
 
 class ItemStateUpdateTrigger(Trigger):
     def __init__(self, itemName, state=None, triggerName=None):
-        triggerName = triggerName or uuid.uuid1().hex
+        if triggerName is None:
+            triggerName = uuid.uuid1().hex
+        else:
+            triggerName = "{}-{}".format(triggerName, uuid.uuid1().hex)
         config = { "itemName": itemName }
         if state is not None:
             config["state"] = state
@@ -50,7 +55,10 @@ class ItemStateUpdateTrigger(Trigger):
 
 class ItemStateChangeTrigger(Trigger):
     def __init__(self, itemName, previousState=None, state=None, triggerName=None):
-        triggerName = triggerName or uuid.uuid1().hex
+        if triggerName is None:
+            triggerName = uuid.uuid1().hex
+        else:
+            triggerName = "{}-{}".format(triggerName, uuid.uuid1().hex)
         config = { "itemName": itemName }
         if state is not None:
             config["state"] = state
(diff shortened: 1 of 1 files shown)


openhab-5iver:
@besynnerlig, @mjcumming, @ScottKinSF , I've tested this pretty well and all of my rules loaded properly, but I prefer at least a second pair of eyes on anything in core.rules or core.triggers. Could one of you please take a look and try this out?

@tavalin, this should resolve your issue, so maybe you could test it out too?

CrazyIvan359:
@openhab-5iver so I built my dev instance finally... test did not go well though:
```
20:44:13.634 [INFO ] [smarthome.event.RuleStatusInfoEvent  ] -
    732a2e3d-7e28-49b9-9207-56c857498039 updated: UNINITIALIZED (INVALID_RULE): Validation
    of rule 732a2e3d-7e28-49b9-9207-56c857498039 has failed! Invalid module
    uid: Item-teststringitem-changed-to-'netService:audio?service=spotify'. It is null or not
    fit to the pattern: [A-Za-z0-9_-]*
```

mjcumming:
@openhab-5iver I might have sometime over the coming days - My install is 2.4 or do I need to test against 2.5M1?

openhab-5iver:
> My install is 2.4 or do I need to test against 2.5M1?

The entire repo is compatible with 2.4, but the custom module handlers (like StartupTrigger) won't work unless you are on a snapshot or 2.5M2 (when it comes out). If you already have everything else up to date, just replace your triggers.py with [this](https://raw.githubusercontent.com/openhab-scripters/openhab-helper-libraries/a43414e3a6ebcc90d984bddca350083944f894e6/Core/automation/lib/python/core/triggers.py).

tavalin:
@openhab-5iver works for me. I've now ported over my first 2 rules to jsr223 👍

More to come in the future...

openhab-5iver:
Thank you!

PR #168, opened 2019-06-21
Add PIN code support for ideAlarm

### LINKED ISSUE #47: [Feature Idea] Logical link @and for triggering rules
Hello, for quite some time i thought about logical and links for triggering rules.

OR-links are already possible with successive @when decorators

```
@when("Item Item1 changed")
@when("Item Item2 changed")
...
```

but how about a corresponding @and decorator to a @when?

As one of many examples i have a rule that should notify me when a door or window was opened while i am on holiday. It triggers every time a door or window is opened:
```
@rule("rNotificationDoorWindowOpenedDuringHoliday")
@when("Descendent of gWindowsDoors changed to OPEN")
def rNotificationDooWindowrOpenedDuringHoliday(event):
	if items.Sw_sRules_HolidayMode == ON:
		# some warning
```

But i would prefer it would only trigger when i am away, because this would not flood the log that much:
```
@rule("rNotificationDoorWindowOpenedDuringHoliday")
@when("Descendent of gWindowsDoors changed to OPEN")
@and("Item Sw_sRules_HolidayMode  == ON")
def rNotificationDooWindowrOpenedDuringHoliday(event):
	# some warning
```

Whats your opinion? Thanks a lot :)

### LINKED ISSUE #74: [Feature Request] Add PIN code support for ideAlarm
It would be good to have the ability to enter a PIN code to disarm the alarm.

I already implemented it locally, I'll try to make a PR once ideAlarm is fully migrated.

fastlorenzo:
Fixes #74


openhab-5iver:
I don't see any issues in the code, but don't have ideAlarm setup to test. @besynnerlig, can you please review and test?

We should also add something to the documentation.

Community/ideAlarm/automation/lib/python/community/idealarm/__init__.py
```diff
@@ -304,16 +305,23 @@ def onToggleSwitch(self, itemName):
         Called whenever an alarm arming mode toggle switch has been switched.
         '''
         newArmingMode = None
+        pinCode = getItemValue(itemName, 0)
+        if pinCode == 0:
```
openhab-5iver: Wouldn't this make a PIN code a requirement for arming/disarming? I think this can be removed.

Community/ideAlarm/automation/lib/python/configuration.py.example
```diff
@@ -38,7 +38,9 @@ idealarm_configuration = {
             'armAwayToggleSwitch': 'Toggle_Z1_Armed_Away',
             'armHomeToggleSwitch': 'Toggle_Z1_Armed_Home',
             'mainZone': True,
-            'canArmWithTrippedSensors': False
+            'canArmWithTrippedSensors': False,
+            'pinCodeEntry': 'Z1_PinCode',
+            'pinCode': 1111
```
openhab-5iver: I suggest setting the default PIN to None.

fastlorenzo: Agreed, I'll make the change

Community/ideAlarm/automation/lib/python/community/idealarm/__init__.py
fastlorenzo: I need to slightly change it indeed, also need to check if self.pinCode is None (pinCode not enabled)

fastlorenzo:
Changes made. I think we should update the doc as well. The armed home and armed away items needs to be Number and not Switch types.

besynnerlig:
> I don't have ideAlarm setup to test. @besynnerlig, can you please review and test?

Sure, I'll do that :)

besynnerlig:
Thank you for your contribution to the repo.

> Fixes #47

That must be wrong. I cant see what this to do with #47 

Except for the title there is not much of a pull request description. The purpose of this Pull Request is quite obvious but can you provide a brief description or summary of the new feature?

I'll have a look at the code within a few days.

Community/ideAlarm/automation/lib/python/community/idealarm/__init__.py
```diff
@@ -304,16 +305,21 @@ def onToggleSwitch(self, itemName):
         Called whenever an alarm arming mode toggle switch has been switched.
         '''
         newArmingMode = None
+        pinCode = getItemValue(itemName, 0)
         if itemName == self.armAwayToggleSwitch:
             if self.getArmingMode() in [ARMINGMODE['DISARMED']]:
                 newArmingMode = ARMINGMODE['ARMED_AWAY']
-            else:
+            elif pinCode == self.pinCode or self.pinCode == None:
```
openhab-5iver: While this might work, the proper way to compare to None is by using 'is'...

    elif pinCode == self.pinCode or self.pinCode is None:

Community/ideAlarm/automation/lib/python/community/idealarm/__init__.py
```diff
@@ -304,16 +305,21 @@ def onToggleSwitch(self, itemName):
...
                 newArmingMode = ARMINGMODE['DISARMED']
+            else:
+                self.log.info(u"Invalid pin code provided to disarm zone [{}]".format(self.name.decode('utf8')))
         else:
             if self.getArmingMode() in [ARMINGMODE['DISARMED']]:
                 newArmingMode = ARMINGMODE['ARMED_HOME']
-            else:
+            elif pinCode == self.pinCode or self.pinCode == None:
```
openhab-5iver: Same as above...

    elif pinCode == self.pinCode or self.pinCode is None:

Community/ideAlarm/automation/lib/python/community/idealarm/__init__.py
fastlorenzo: Fixed

Community/ideAlarm/automation/lib/python/community/idealarm/__init__.py
fastlorenzo: fixed as well

fastlorenzo:
Summary of the new feature: a PIN code can be used to disarm the alarm when armed home/away.
The PIN code needs to be set in the configuration.py file and must be a digit.
If you do not wish to use the PIN code feature, the pinCode property can be set to None.

When using the pinCode feature, the zone armAwayToggleSwitch and armHomeToggleSwitch needs to be set to a Number and not a Switch type.

fastlorenzo:
> 
> Wouldn't the approach that I suggest above not only be much easier but also more flexible do you think? It won't require any breaking changes to ideAlarm at all.

This could indeed work but I took the approach of changing the toggle switches for security reasons.
When you use the pinCode, you cannot disarm the alarm without knowing the pin code which is hard coded in the python script, which means even if you have access to the event bus of OH you cannot disarm the alarm.

>
> You haven't added a separate alarm event that can be triggered whenever a faulty PIN code has been entered hence what will follow when that happens is hard coded (a log entry). Maybe some users would prefer to have a light to turn on, a message be sent etc.

True, I think this could be a good addition for ideAlarm to have an item which contains an error message. A specific item with invalid pin code might work as well.

fastlorenzo:
I wanted to update the documentation, but don't know how to generate the html from the rst.txt files. Any help?

CrazyIvan359:
@fastlorenzo you should be adding or editing documentation in the `.rst` files in the `/Sphinx` directory. I believe Scott is pushing the changes to the output files separately so that PRs are more readable since rebuilding the pages can sometimes modify up to 100 or more files.

If you want to build the pages so you can see them for proofing purposes, there are instructions [here](https://openhab-scripters.github.io/openhab-helper-libraries/Contributing/Writing%20Docs.html), just don't commit any changes to files in the `/docs` directory.

besynnerlig:
> > Wouldn't the approach that I suggest above not only be much easier but also more flexible do you think? It won't require any breaking changes to ideAlarm at all.
> 
> This could indeed work but I took the approach of changing the toggle switches for security reasons.
> When you use the pinCode, you cannot disarm the alarm without knowing the pin code which is hard coded in the python script, which means even if you have access to the event bus of OH you cannot disarm the alarm.

You could. Frankly, if an intruder can access the Event Bus, e.g use a script to send a given command to the specified Item there are numerous ways to break the functionality of ideAlarm.
for example `sendCommand("Z1_Arming_Mode", 0)`

> Regarding the breaking change, I agree this is one if the pinCode is set to something else than None, but if it is set to None (which is the case by default), the script works as previously (using Switch items).

No it won't.
For example trying to get the integer value of a switch like in your statement `pinCode = getItemValue(itemName, 0)` will throw `AttributeError: 'org.eclipse.smarthome.core.library.types.OnOffType' object has no attribute 'intValue'`

Even if you adapt your code in some way you must remember that the users have custom scripts that expect the arming mode toggle switches to be of the type switch. If we shall change the item type for the arming mode toggling switches we need to gain something quite valuable to do so. We also need to rename the items to not hold the word "switch" in their names because it's confusing. Unfortunately I can't see that we gain security or anything else by doing this.

> > You haven't added a separate alarm event that can be triggered whenever a faulty PIN code has been entered hence what will follow when that happens is hard coded (a log entry). Maybe some users would prefer to have a light to turn on, a message be sent etc.
> 
> True, I think this could be a good addition for ideAlarm to have an item which contains an error message. A specific item with invalid pin code might work as well.

It's not what I meant. ideAlarm has [Event Helpers](https://openhab-scripters.github.io/openhab-helper-libraries/Python/Community/ideAlarm/Event%20Helpers.html) for specific events. That's the correct place to create your own error messages. If your PR would be approved we'd need another event helper: `onPinEntered(zone, pincode)` . Just letting ideAlarm log an entry when a faulty PIN is entered is not enough. Having that would also allow for defining and handling multiple PIN codes. But all this and more is very easy to implement using the simple script that I suggested above.

I still can't see any benefits using the suggested PR over using the simple script example that I suggested above. If you don't agree with me, please convince me by giving me more details in how the security is improved and please let me know how a user can customize what will happen when a PIN code (correct or faulty) has been entered.

fastlorenzo:
I'll try your approach and see if it can work, it can indeed be easier to implement than all the breaking changes I wanted to introduce.

> You could. Frankly, if an intruder can access the Event Bus, e.g use a script to send a given command to the specified Item there are numerous ways to break the functionality of ideAlarm.
for example sendCommand("Z1_Arming_Mode", 0)

Good point, it might be interesting to find a solution so only the ideAlarm script can change the state of those items. I'm not sure this is something feasible in OH at the moment.

besynnerlig:
> I'll try your approach and see if it can work, it can indeed be easier to implement than all the breaking changes I wanted to introduce.

Please do that and please don't hesitate to ask if there's something that you can't solve by using the example script.

If you after trying think that the script is working well, we could add the example script to the documentation.

besynnerlig:
> I'll try your approach and see if it can work, it can indeed be easier to implement than all the breaking changes I wanted to introduce.

@fastlorenzo , have you had the time yet to check it out?

openhab-5iver:
Pinging @fastlorenzo... shall we close this?

openhab-5iver:
Closing due to lack of response.

PR #186, opened 2019-07-23
[utils] Rename postUpdateCheckFirst and sendCommandCheckFirst

### LINKED ISSUE #185: [utils] Rename postUpdateCheckFirst and sendCommandCheckFirst
`utils.py` functions `postUpdateCheckFirst` and `sendCommandCheckFirst` do not have names particularly evident of their function. I suggest renaming them to `postUpdateIfDifferent` and `sendCommandIfDifferent` for clarity, and creating aliases for the old names to maintain backwards compatibility.

CrazyIvan359:
Updates names of `postUpdateCheckFirst` and `sendCommandCheckFirst` to `postUpdateIfDifferent` and `sendCommandIfDifferent`. The old names are not very clear about the use of the function. The original names have been preserved as aliases for backwards compatibility.

Fixes #185 

Signed-off-by: Mike Murton <6764025+CrazyIvan359@users.noreply.github.com>

### DIFF
--- a/Core/automation/lib/python/core/utils.py
+++ b/Core/automation/lib/python/core/utils.py
@@ -93,7 +93,7 @@ def postUpdate(itemName, newValue):
     item = scope.itemRegistry.getItem(itemName) if isinstance(itemName, basestring) else itemName
     scope.events.postUpdate(item, newValue)
 
-def postUpdateIfDifferent(itemName, newValue, sendACommand=False, floatPrecision=None):
+def post_update_if_different(itemName, newValue, sendACommand=False, floatPrecision=None):
     '''
     newValue must be of a type supported by the item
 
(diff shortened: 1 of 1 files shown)


openhab-5iver:
I almost merged, but had a thought. What if we used this as an opportunity to snake_case them?

CrazyIvan359:
Didn't we decide we should be going the other way? The goal is to merge most of these libraries into oh core or an addon, oh is all camelCase.

openhab-5iver:
> Didn't we decide we should be going the other way? 

No... if anything, everything should be converted to snake_case to help prevent name conflicts.

CrazyIvan359:
While I agree with that, wouldn't that be bizarre after integration with oh? Having oh in camelCase and our addon in snake_case.

openhab-5iver:
Why do you think it would be bizarre? Once the Scripting API is in place, everything will be in mixedCase, except for anything still left in the libraries. I think it would help to differentiate what you are using.

CrazyIvan359:
Oh I see, you don't mean to avoid naming conflicts with OH as it is now, you mean between the HLs as they are now and the forthcoming API!

openhab-5iver:
Correct. There could potentially be other things in OH or Java too, but the Scripting API would be a major source of name conflicts.

CrazyIvan359:
Yes, now I understand. I will update the names

besynnerlig:
Even though the original names have been preserved as aliases for backwards compatibility, maybe we should consider to rework all the included libraries and examples (e.g. only ideAlarm a.f.a.i.k) to use the new names.

openhab-5iver:
Once this is merged, I'll update any usage in the repo with new function names...

https://github.com/openhab-scripters/openhab-helper-libraries/search?q=postupdatecheckfirst&unscoped_q=postupdatecheckfirst
https://github.com/openhab-scripters/openhab-helper-libraries/search?q=sendcommandcheckfirst&unscoped_q=sendcommandcheckfirst

CrazyIvan359:
@openhab-5iver shouldn't that renaming happen in this PR?

openhab-5iver:
Yes, but I didn't want to make work for you. But you're going to need to rebase anyhow, so might as well get it in.

CrazyIvan359:
Rebase seems like a lot of trouble for a small PR like this. Just make the changes to master and close this PR with a reference to the commit instead.

PR #206, opened 2019-08-09
[metadata] Added concept of key hierarchies

openhab-5iver:
Previously, the access to metadata was flat and you could only navigate the top level. This PR allows you to now access and modify nested dicts with the metadata's configuration. Details in docstrings.


CrazyIvan359:
This doesn't contain any changes to metadata? All I see are a bunch of changes to "But How Do I?" and a comment in metadata.py

PR #213, opened 2019-08-11
Use simplified installation instructions for Python

CrazyIvan359:
This PR updates the documentation to use the `lib/ext/` dir and `registry` file to make installing Python/Jython less complicated.

Signed-off-by: Michael Murton <6764025+CrazyIvan359@users.noreply.github.com>

### DIFF
--- a/Sphinx/Getting Started/Docker.rst
+++ b/Sphinx/Getting Started/Docker.rst
@@ -17,7 +17,7 @@ Docker
                 
             .. code-block:: bash
 
-                wget https://github.com/OH-Jython-Scripters/openhab2-jython/archive/master.zip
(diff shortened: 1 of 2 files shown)


openhab-5iver:
Using /runtime/lib/ext/ and the Jython registry works, but I prefer the current installation instructions. This method simplifies things slighty by using the registry instead of the EXTRA_JAVA_OPTS, but [the runtime directory is wiped](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/update#L295) during an OH upgrade, so you'd need to reinstall Jython every time. This could be remedied if a change was allowed to the update scripts, but I'd rather just get the addon working.

PR #219, opened 2019-08-20
Patch 1

volfan6415:
Update rules.js  …
fix two bugs with new Simple Rule object creation

add support for adding conditions to rule defintions

Conditions.js

this simplifies the code for ItemStateCondition as well as fixes an error where generic compare condition would not create valid condition types


openhab-5iver:
@volfan6415, sorry for the delay... my time has been very limited. Could you please split this PR into two... one for rules.js and one for conditions.js?

volfan6415:
Done.  Created 286 and 287 this pull can be closed.

I broke the split based on functionality vs the files.  There were changes in rules.js which need to be made for the conditions to work properly.

Issue #220, opened 2019-08-30
Submit the Gatekeeper Design Pattern to the library


rkoshak:
**Is your feature request related to a problem? Please describe.**
Implement the Gatekeeper design pattern which applies to cases where one needs to rate limit or enforce a pause after executing one command before issuing another command. This can be useful in technologies like RF433 or the say action.

**Describe the solution you'd like**
Provide this as a library capability so users do not need to copy/paste/implement this common requirement on their own.

**Describe alternatives you've considered**
Forum posting, adding it to the Design Patterns section of the docs. But I've implemented the DP in a generic and reusable way which will let users simply reuse the code instead of needing to copy and use it themselves. 

**Additional context**
See https://community.openhab.org/t/design-pattern-gate-keeper/36483 for the initial impelmentation, a PR is to follow shortly with a documented version with tests that more closely follows the Python style guide.

5iver:
#232

PR #232, opened 2019-09-03
Gatekeeper DP

rkoshak:
Initial submission of the Gatekeeper DP for inclusion in the Helper Libraries.

It's implemented as a class to support the option to have more than one gatekeeper active in a single set of Rules (e.g. one gatekeeper to manage commands to a 433 device and another to manage calls to say). 

Solves #220


Community/Gatekeeper/automation/lib/python/community/gatekeeper/gatekeeper.py
```diff
@@ -0,0 +1,95 @@
...
+        allowed to execute 750 msec after the command returns.
+
+        Args:
+            - pause: Time in milliseconds to wait until the next command will be
+            allowed to execute.
+            - command: Lambda or function to call to execute the command.
+        """
+        self.commands.put((pause, command))
+        if self.timer is None or self.timer.hasTerminated():
+            self.__proc_command__()
```
CrazyIvan359: This makes `add_command` block until the function ran if the queue was empty. You should instead start the timer with a delay of 0 so that a different thread runs `__proc_command__`.

Also wrapping names in double underscores is reserved for python internals and should not be used. For module/class internals you should preceed them with a single underscore like `_proc_command`.

rkoshak: I definitely do not want to force the first command executed when the queue is empty to have to wait some amount of time before it can run. It should run immediately if it is allowed to run which is why I purposefully don't create a Timer in that case. The gatekeeper's purpose is to enforce a delay after a command runs and should not unnecassarily add delays before a command runs. In short, this was a deliberate design decision.

Now we can argue whether add_command should block or not on that first command. I don't really see why that would be a problem but I can be convinced it is given a use case where that would make a difference. I just can't think of one off hand. But if it can be run immediately, it must run immediately without delay. So I'm not certain a Timer is the right approach. 

I'll fix the double underscores for sure.

CrazyIvan359: I think I misunderstood the use case when I looked at it before. Looking at it again I do see that it makes sense this way.

rkoshak:
I realized that this class can be used to implement the Cascading Timers DP, but only if there is a way to clear out the queue (i.e. cancel) so I added that feature and tests to make sure it works.

rkoshak:
This PR is ready for review

rkoshak:

PR #233, opened 2019-09-03
Timer Manager

rkoshak:
A class that manages the case where a user needs to keep track of one Timer per Item. The class supports creation, cancelling, and performing different actions when the Timer expires or when the Timer already exists. I chose a class because I actually have a case in my Rules where I need to keep track of two different timers for the same Items.


rkoshak:
Question about dependencies. I'm working an implementation of the Generic Presence Detection tutorial which would benefit from using this library. Should Community modules and scripts be self contained or can they depend upon each other (with copious documentation of course)?

rkoshak:
This PR is ready for review

rkoshak:

PR #235, opened 2019-09-09
Update date.py

stagf15:
First, I found an issue in the to_joda_datetime() function where passing it a python datetime.datetime object would throw an error: 
IllegalArgumentException: java.lang.IllegalArgumentException: The datetime zone id 'GMT-05:00' is not recognised
I believe this is because java ZoneOffset.getId is expecting just "-05:00" without the 'GMT', and is "incompatible" with the python zone ID.  I modified the code slightly and replace "GMT" with "", and it clears up the problem for me (line 305-308), and shouldn't cause any side effects.

Also, I added a millis() function to get millis past unix epoch, a joda_now() function to simplify getting the current joda datetime string or object, an elapsed function to get various formats of elapsed time between two datetime objects or joda datetime strings, and a str_to_python_datetime() function to allow passing a joda datetime string into all of the other conversion functions within this library.

Note: I did this because I have all of these functions in a separate personal library and I wanted to combine them, and I also wanted to learn about github and the fork/pull process.  Please let me know if I'm breaking some kind of etiquette rules here! Thanks.

I know there's a change forthcoming on the human_readable_seconds() function to convert the argument to a integer, and I'm interested to see if my fork gets that update or if a fork gets orphaned at the point where it is "forked".  (Lack of github knowledge on my part).

### DIFF
--- a/Core/automation/lib/python/core/date.py
+++ b/Core/automation/lib/python/core/date.py
@@ -57,7 +57,7 @@ __all__ = [
     "format_date",
     "days_between", "hours_between", "minutes_between", "seconds_between",
     "to_java_zoneddatetime", "to_java_calendar", "to_python_datetime",
-    "to_joda_datetime", "str_to_python_datetime", "millis", "joda_now", "elapsed"
+    "to_joda_datetime", "millis", "joda_now", "elapsed"
 ]
 
 
--- a/Core/automation/lib/python/core/date.py
+++ b/Core/automation/lib/python/core/date.py
@@ -414,7 +414,7 @@ def elapsed(start, end=joda_now(), format='digital'):
         start:  Start time in any datetime type or joda zdtt string (Required)
         end:    End time, optional. If not included the current time will be used
         format: 'digital' - Returns string '(-)HH:MM:SS' or '(-)Dd HH:MM:SS' [DEFAULT]
-                'text'    - Returns human readable string '2 days, 3 hours, 45 minutes'
+                'text'    - Returns human readable string '2 days, 3 hours and 45 minutes'
                 'seconds' - Returns the seconds as a float (with ms precision)
     
     Returns:
@@ -442,7 +442,7 @@ def elapsed(start, end=joda_now(), format='digital'):
     end_pdt = to_python_datetime(end)
     td = (end_pdt - start_pdt)          # Total time delay (as timedelta)
 
-    ts = e_seconds = td.total_seconds() # Retains ms precision
+    ts = e_seconds = td.total_seconds() # Retains ms precision (total seconds)
     sign = '-' if (ts < 0) else ''      # Get the sign of 'ts'          
     ts = int(abs(ts))                   # Make 'ts' a positive integer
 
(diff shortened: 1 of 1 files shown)


CrazyIvan359:
I think `joda_now` should return a Joda Time object, not a string by default. Returning a string is not what is implied by that function name and would be a less common use case.

In regards to `str_to_python_datetime` it should used Java DateTime as that is preferred for this module for future compatibility and inclusion into openHAB.

As for github, learning git on the command line is a must, there is so much more functionality that you just can't get from WebUI. You will need to keep your fork up to date manually, but git makes it easy. Github actually provides a tutorial on how to do this, Google is your friend.

stagf15:
Thank you for the comment, I will make your suggested changes tomorrow. I’ll have to do some studying to figure out how to do the conversion from string to Java ZonedDateTime. And I’ll probably make a new function called str_to_java_zoneddatetime().

BTW, The only reason I made the joda_now() function return a string by default is because it seems like almost everything must be turned into a string to pass into an item update/command. Is there a way to actually pass objects instead of strings into all of the various item types? It just feels wrong converting everything to a string.

openhab-5iver:
> I will make your suggested changes tomorrow

Best not to make any changes until I have done a review, or you may be wasting your time. Unfortunately, I don't have the time for it ATM and there are a few other PRs in queue before yours.

> I know there's a change forthcoming on the human_readable_seconds() function to convert the argument to a integer, and I'm interested to see if my fork gets that update or if a fork gets orphaned at the point where it is "forked".

You will get to learn how to do a rebase after I get it committed. Basically, your fork is a snapshot in time since it diverged from master, but you've also made changes. Your fork will need to be updated with changes in master before your commit can be merged due to other changes made in the same file. So, when you pull in the changes, you'll need to interleave your changes and the ones from master. When learning, always make a backup of your local repo. That way you can experiment and restore from backup when things get messy... and they will! If you haven't already, setup VS Code. It has some basic UI for git commands, has a terminal for doing things from the commandline, and there are some very helpful extensions.

> it should used Java DateTime as that is preferred for this module

Says who?! These helper libraries need to support Joda, Java and Python datetimes, with Joda at the top of the list until OH has removed it completely, which won't be until OH3.

CrazyIvan359:
I seem to recall that it's possible to pass a string to the constructor. If that's the case you should do the check for `basestring` in `to_java_zoneddatetime` and return immediately like the other conversations in that function. You should then use `to_java_zoneddatetime` in your function and remove the checks for `basestring` in the other conversions since they won't be needed.

While I understand and agree with your reasoning I think it should return a Joda Time object. The goal of this module to to help manipulate dates and times, when sending commands to items you can use the format function to get the time as a string.

stagf15:
@CrazyIvan369

> I seem to recall that it's possible to pass a string to the constructor. If that's the case you should do the check for basestring in to_java_zoneddatetime and return immediately like the other conversations in that function. You should then use to_java_zoneddatetime in your function and remove the checks for basestring in the other conversions since they won't be needed.

This is done, great idea. Thank you!

openhab-5iver:
I'm closing this due to inactivity. The changes were never rebased, as requested, but there are also too many changes going into a single PR. I will submit a PR for the fix for the bug in to_joda_datetime.

PR #237, opened 2019-09-12
Initial submission of a Countdown Timer implementation

rkoshak:
This is a module that has a class to implement a countdown timer. It creates a timer that loops every second, updating a passed in Item with the amount of time left on the Timer so people can see how much time is left on their sitemaps and such. 

A Number Item or a String Item can be passed to receive the time left updates. If a Number, the number of seconds (rounded to the nearest second) remaining updates the Item. If a String Item, the result of str on the datetime.timedelta Item get's returned, again rounded to the nearest second. The timer itself though will expire on the scheduled millisecond 


rkoshak:
This PR is ready for review

rkoshak:

PR #255, opened 2019-09-16
Rate Limiting.

rkoshak:
This is an implementation of non-blocking rate limiting implementation. When you call "run" it will ignore any calls to it that take place before the latch time expires.

I chose a Class in this case because a user may need more than one limiter going at any given time.

NOTE: I had to rename everything as I'm clearly tired.


rkoshak:
This PR is ready for review

rkoshak:

PR #259, opened 2019-09-20
Drop in replacement for Expire Binding

rkoshak:
This is a script that acts as a drop in replacement for the Expire1 binding. Because the binding config gets stored as metadata one does not even need to change their Items.

I don't intend this to be a permanent replacement but to be a bridge to whatever comes along to replace it or if the Expire1 binding is no longer supported without a suitable replacement in place. It can also be used by those right now who want to eliminate their 1.x version bindings and retain the functionality.

I mainly wrote it as a challenge to see if I can do it and I'm making an effort of move off of all the 1.x version bindings, even those that don't yet have a 2.x version replacement. Perhaps others will find this useful.

Based on the tests I've done so far it is feature complete and exhibits the same behaviors as the Expire binding except for the fact that the script needs to be reloaded when adding or removing Items to the expire so the triggers can be regenerated. Without a way to trigger a Rule when Item metadata changes I'm not sure if I can ever overcome this limitation.

### DIFF
--- a/Community/Expire/automation/jsr223/python/community/expire/expire.py
+++ b/Community/Expire/automation/jsr223/python/community/expire/expire.py
@@ -1,0 +1,193 @@
+"""
+Author: Rich Koshak
+
+Implements a drop in replacement for the Expire1 binding using Rules.
+
+Requirements:
+    - Uninstall the Expire1 binding before use
+
+Limitations:
+    - The expire config metadata string must include the default units when
+    using with Items that are defined with Units of Measure.
+    - Adding new Items or changing Items with an expire metadata config requires
+    a reload of this script to regenerate the rule triggers.
+
+License
+=======
+Copyright (c) contributors to the openHAB Scripters project
+"""
+from core.rules import rule
+from core.triggers import when
+from core.metadata import get_value
+import re
+from datetime import timedelta
+from core.actions import ScriptExecution
+from org.joda.time import DateTime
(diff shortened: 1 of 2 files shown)


rkoshak:
Once this PR is accepted, I'd like to extract the duration parsing into it's own library, perhaps in core.util. I've found other places where it's useful.

bwosborne2:
This 1.x dependency needs to be remove, I believe.

from org.openhab.core.library.items import StringItem

OH cannot find org.openhab.core.library.items

rkoshak:
This PR is ready for review

CrazyIvan359:
Sorry gents, I'm gonna have to disturb the dust on this one. Just a minor change though.
@rkoshak I'll be making a PR against your branch in a few minutes.

Stumbled on a bug in removing the old rule, using `rules` silently fails. Bringing in the `ruleRegistry` and using that works fine.

rkoshak:
I'll wait for Scott to give the go ahead before I merge the PR in case he has already started.

jimtng:
what's the hold up?

rkoshak:
Still awaiting review. I suspect reviewing community library submissions has been and continues to be a lower priority. Or the maintainers have decided to go a different direction. Any further development on these PRs is halted until they get reviewed. You can find CrazyIvan359's updates at https://github.com/rkoshak/openhab-helper-libraries/pulls

rkoshak:

PR #260, opened 2019-09-23
Corrected link to jython download page

Urmel:


openhab-5iver:
That is not the correct link for the 2.7.0 standalone jar... but this is... search.maven.org/remotecontent?filepath=org/python/jython-standalone/2.7.0/jython-standalone-2.7.0.jar.

Urmel:
I proposed the download site so that the users understand that this is the official JAR. The direct link goes to a maven site.
Thanks for building these libs!!

openhab-5iver:
> I proposed the download site so that the users understand that this is the official JAR.

I like the thought, but since it is not directly linking to 2.7.0, it could cause people to download a newer version. I don't recommend Jython 2.7.1 due to some bugs and some compatibility issues with OH, I have workarounds in the helper libraries and use Jython 2.7.2a+ in my production and test environments. So, it is not really an issue and I could be swayed to change my mind, but for now I think it is best to stick with 2.7.0. Our options are to use a direct link or a link to the general downloads page with more detailed instructions on how to get 2.7.0 (under Previous Versions). I vote for the former, until 2.7.2 is released.

> Thanks for building these libs!!

I am the current maintainer, but not the original contributor, so I can only take a portion of the credit. This repo is also community driven, so contributions are welcome from everyone!

I won't stop you from [buying me a beer](https://www.paypal.me/ScottRushworth) though :wink:!

Urmel:
Okokok. You got me. Changed the link.
Sorry - took some time.

openhab-5iver:
Thank you for the :beer:!!!!!

Issue #267, opened 2019-10-15
Use a JS module system for JS rules


jpg0:
**Is your feature request related to a problem? Please describe.**
I find the custom library loading code frustrating because:
- It's pretty long boilerplate to import a library
- Imports pollute the global namespace
- Writing libraries or importing them is unfamiliar to JS coders
- Not compatible with existing libraries

**Describe the solution you'd like**
- CommonJS or AMD implementation for libraries

**Describe alternatives you've considered**
- I have NOT looked into the aboves' compatibility with OH, but assumed that it's possible one way or another :)

openhab-5iver:
I'm pretty certain these will not be possible, as we need to stay native Nashorn for javax.script. Even if it was possible, I do not feel it would be worthwhile implementing a module system, since JDK9/ES6 is very close and will change everything.

jpg0:
So I was thinking of just something basic which meant that you could use a loader syntax. To see if it was possible, I just created something basic so that you can write code like this:


         
In script:
```
load(Java.type("java.lang.System").getenv("OPENHAB_CONF")+'/automation/lib/javascript/personal/loader.js');

var foo = require('foo');
foo.bar();
```
In library:

```
exports.bar = function(){ ... }
```
I can create a PR to add the loader.js file if you like / are interested?
ps.I am also hoping for JDK9 but from the actual movement I see in OH I'm not confident it'll happen for a while.

Martin-Stangl:
@jpg0, I have a question for clarification: Is it correct to assume that the library file containing the code `exports.bar = function(){ ... }` is called foo.js? And is it also located in /automation/lib/javascript/personal/?
Or is there some magic going on in loader.js that maps foo somehow to something?

In general using modules is a great idea. But it would be a major rewrite and keeping backwards compatibility to the current libraries will be quite some effort - probably more than it is worth..

jpg0:
@openhab-5iver I'm glad that you're more optimistic than I am! It was exactly that issue (and the lack of progress on it + it's dependencies) that made me think it will be a while!

@Martin-Stangl yes that is all correct. Loader is not magic at all, I just made some minor modifications to an existing pure-JS one: https://gist.github.com/bspaulding/1386829. Saying that, it does not currently prevent libraries polluting the global namespace, but well-behaved ones should work fine (this may be possible using loadWithNewGlobal in Nashorn, but I haven't looked into it).

The reason that I am suggesting doing this now is that:
[...]

PR #266, opened 2019-10-13
Create items.js: creation/removal of items

jpg0:
Port dynamic items creation/removal from python core libs

PR #270, opened 2019-10-21
Added initial CommonJS support

jpg0:
PR as discussed in #267 

This implements a CommonJS module system for loading libraries.

- requiring and exporting work just like you would expect
- this implementation DOES provide isolation from libraries' adding to the global namespace, by loading each library into a script engine with it's own context, then referencing the exports inside it
- I have not added a version of log.js (as in #268), although this is trivial to port
- I have removed the requirement for defining 'me' as it's incompatible here, and simply used the uuid

I do believe that pulling the loading engine up into OH code would be better, but agree that we should wait for JDK9/ES6 before we do this (although if it really drags on, I may do it). Either way, this PR means that all library code written now would be compatible with that future.

### DIFF
--- a/Core/automation/lib/javascript/core/PersistenceExtensions.js
+++ b/Core/automation/lib/javascript/core/PersistenceExtensions.js
@@ -7,11 +7,18 @@
  */
 'use strict';
 
-var PersistenceExtensions	= Java.type("org.eclipse.smarthome.model.persistence.extensions.PersistenceExtensions");
+// START: Backward Compatibility Header
+load(Java.type("java.lang.System").getenv("OPENHAB_CONF")+'/automation/lib/javascript/core/init.js');
+// END: Backward Compatibility Header
+
+
 
 //Simplifies spelling for rules.
 (function(context) {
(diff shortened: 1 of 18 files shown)


jpg0:
Oh, and one other thing as mentioned elsewhere: this really shows that utils.js should be broken up into logical parts. It seems to cover a little bit of everything. For example one of the first things that I'd do is move `getItem`, `sendCommand`, `postUpdate` to items.js (#266)

jpg0:
After looking at this, I have discovered that there is a problem:
- Isolation happens by loading scripts into a new context, in a new engine
- This generally works fine
- However `scriptExtension.importPreset` isn't present in these contexts
- I can however pre-load all the properties from these and supply them into sub-contexts via synthetic modules (then they use via require('RuleSupport') for example).
- This appears to not work well however, causing issues calling functions back in the parent context and marshalling arguments through
- Ultimately, I think providing isolation this way is too complex and needs to be done in the above OH layer.
I'll close this PR and open a new one without isolation.

PR #279, opened 2019-11-05
ideAlarm: Added attribute to auto reset zone after alert

Mr-Eskildsen:
Added an attribute 'autoReset' to each zone in ideAlarm. When false, functionality will be unchanged. If true the zone will automatically change to idle (without the need to acknowledge the previous alarm). 

In my case I will use it for outdoor PIR sensors which should turn on a light upon movement. It should thus rearm and be ready to trigger another movement.


besynnerlig:
Thanks. I'll have a look at it.

Community/ideAlarm/automation/lib/python/community/idealarm/__init__.py
```diff
@@ -166,6 +166,7 @@ def __init__(self, parent, zoneNumber, cfg):
         self.name = cfg['name']
         self.armAwayToggleSwitch = cfg['armAwayToggleSwitch']
         self.armHomeToggleSwitch = cfg['armHomeToggleSwitch']
+        self.autoResetAfterAlert = cfg['autoReset']
```
besynnerlig: Since this is a new configuration key, I'd prefer if we check for the existance of the key instead of assuming it's already in the configuration. (E.g. if key in dict: ) I'd prefer if this new feature won't be a "breaking change", that is, if configuration key is not added, things will just keep working as they did before.

Community/ideAlarm/automation/lib/python/configuration.py.example
```diff
@@ -38,7 +38,8 @@ idealarm_configuration = {
             'armAwayToggleSwitch': 'Toggle_Z1_Armed_Away',
             'armHomeToggleSwitch': 'Toggle_Z1_Armed_Home',
             'mainZone': True,
-            'canArmWithTrippedSensors': False
+            'canArmWithTrippedSensors': False,
+			'autoReset': False
```
besynnerlig: Tab indentation character used here and below. I believe it should be 4 space characters as elsewhere in the file.

Mr-Eskildsen:
Thanks for the suggestions in the review. This should now be corrected in the pull request

Community/ideAlarm/automation/lib/python/community/idealarm/__init__.py
openhab-5iver: Is there a default value, so that you could use something like this?

    self.autoResetAfterAlert = cfg.get('autoReset', default_value)

Mr-Eskildsen: I have implemented it with a check for the key. Your suggestion seems a little bit more compact, but the result will be the same?

openhab-5iver: Exactly the same

Community/ideAlarm/automation/lib/python/personal/idealarm/custom_template_verbose.py.example
```diff
@@ -162,3 +162,10 @@ def onNagTimer(zone, nagSensors):
         tts(msg, PRIO['HIGH'])
     elif zone.zoneNumber == 2 and scope.itemRegistry.getItem('Z2_Block_Nag').state != scope.ON:
         tts(msg, PRIO['MODERATE'])
+
+
+def onZoneResetAfterAlert(zone, armingMode, zoneStatus):
+    '''
+    This function will be called when the zone is status is automatic reset after being tripped.
+    '''
+    log.info("onZoneResetAfterAlert: {} ---> {}".format(zone.name.decode('utf-8'), kw(ZONESTATUS, zoneStatus)))
```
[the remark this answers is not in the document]
Mr-Eskildsen: Added unicode conversion to the log.info (two places)

openhab-5iver:
Yes, I'll merge, but I just noticed that the commits were never signed... https://www.openhab.org/docs/developer/contributing.html#sign-your-work.

Mr-Eskildsen:
> Yes, I'll merge, but I just noticed that the commits were never signed... https://www.openhab.org/docs/developer/contributing.html#sign-your-work.

Any progress on this?

openhab-5iver:
I only see one commit with a Signed-off-by, but I can make this work when I squash them. Sorry for the delay and thanks for the ping!

PR #281, opened 2019-11-08
UpdateActions.rst to reflect 2.x mail binding action

vzorglub:
UpdateActions.str to reflect new mail binding action
Signed-off-by: Vincent Regaud <vzorglub@gmail.com>

### DIFF
--- a/Sphinx/Guides/Actions.rst
+++ b/Sphinx/Guides/Actions.rst
@@ -170,6 +170,10 @@ Use an Addon/Bundle Action
 
         .. code-block::
 
+            #New Mail Binding Action
+            actions.get("mail", "mail:smtp:mail_thing").sendMail("someone@example.com", "This is the subject", "This is the message")
+            
+            #Deprecated due to new Mail Binding
             from core.actions import Mail
             Mail.sendMail("someone@example.com", "This is the subject", "This is the message")
 


openhab-5iver:
This gets frustrating... but you're learning something new! You added the sign-off into the body of the PR but not into the commit message, which gets stored with the commit and is used to identify the commiter. :slightly_frowning_face:

Let me know before you run off in frustration, because I can do some magic behind the scenes. I'd rather not though, since it is a good learning experience.

vzorglub:
Do it,
I think GitHub is great but I cannot get my head around how it works
I have tried before and it’s just a maze of unnecessary complications

And don’t get me started on trying to use it locally... That is black magic
I just don’t get it

vzorglub:
You added the sign-off into the body of the PR but not into the commit message, which gets stored with the commit and is used to identify the commiter.

Next time hopefully

rkoshak:
> And don’t get me started on trying to use it locally... That is black magic
I just don’t get it

I've actually found that working through VSCode is more self explanatory and intuitive. I wrote up a reply to a thread on the forum somewhere...

https://community.openhab.org/t/dco-check-signing-off-with-github-web-editor-explanation/83330/6

Maybe this will make the workflow a little easier on you. I know it has for me.

PR #282, opened 2019-11-11
Initial check in for Ephemeris Time of Day

rkoshak:
Implements the Time of Day Design Pattern using Ephemeris to define and detect day types (holidays, weekends, etc.). The code supports a different set of times of days based on the type of the day.

### DIFF
--- a/Community/Ephemeris Time of Day/automation/jsr223/python/community/ephem_tod/ephem_tod.py
+++ b/Community/Ephemeris Time of Day/automation/jsr223/python/community/ephem_tod/ephem_tod.py
@@ -62,7 +62,7 @@ def tod_update_items(log):
         time_str = get_key_value(start_time.name, "ToD", "start_time")
 
         # If there is a value, parse it and set the Item to that time for today
-        if time_str:
+        if time_str != {}:
             log.debug("Handling {}".format(start_time))
             time_parts = time_str.split(':')
             num_parts = len(time_parts)
@@ -153,7 +153,7 @@ def get_start_times(log):
         rval = curr
         for grp in [grp for grp in tod_grps if check(grp, type)]:
             value = get_key_value(grp.name, "ToD", key)
-            if value is None:
+            if value == {}:
                 log.error("Group {} doesn't have a key {}!"
                                     .format(grp.name, key))
             elif day_check(value):
@@ -166,7 +166,7 @@ def get_start_times(log):
     start_times = None
 
     start_times = get_group("default",
-                            lambda grp, type: check(grp, type) or not get_key_value(grp.name, "ToD", "type"),
+                            lambda grp, type: check(grp, type) or get_key_value(grp.name, "ToD", "type") == {},
                             start_times,
                             True)
 
@@ -231,7 +231,7 @@ def create_tod_timers(log, start_times):
 
         state = get_key_value(start_time.name, "ToD", "tod_state")
         # If there is no state we can't use this Item.
-        if state is None:
+        if state == {}:
             log.error("{} does not have tod_state metadata!"
                       .format(start_time.name))
 
--- a/Community/Ephemeris Time of Day/automation/jsr223/python/community/ephem_tod/ephem_tod.py
+++ b/Community/Ephemeris Time of Day/automation/jsr223/python/community/ephem_tod/ephem_tod.py
@@ -1,0 +1,436 @@
+"""
+Author: Rich Koshak
+
+Implements the Time of Day Design Pattern using Ephemeris to determine the day
+type so one can have a different set of times of day for different day types.
+
+Requirements:
+    - openHAB 2.5 M4 or later
+    - Ephemeris configured
+
+Limitations:
+    - If using the Astro binding, the Rule that creates the Timers may be
+      triggered more than once 30 seconds after midnight as Astro updates the
+      Items. In testing this doesn't appear to cause any problems.
+    - This script needs to be reloaded if Items are added or removed from
+      tod_group_ephem or metadata changed on any of the Items this script
+      needs to be reloaded.
+
+License
+=======
+Copyright (c) contributors to the openHAB Scripters project
+"""
+
+from core.rules import rule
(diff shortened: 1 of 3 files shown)


rkoshak:
This PR is ready for review

mjcumming:
Also would suggest adding this to reload the config file.

`import configuration
reload (configuration)
from configuration import tod_group_ephem, tod_item_ephem`

rkoshak:
Have you found that in practice? I've had all sorts of timing but I've never had Astro not populate an Item before Rules start running. It's a good test to have there and I should probably test for UNDEF there too because I want this to work with _any_ DateTime source, not just Astro and some bindings can set Items to UNDEF when they go offline. 

I'm just curious if you've actually seen this with Astro. I suppose if you hadn't you would have made the comment. :-)

As for the reload, I'm not yet sure what Scott's policy is on having that in the community scripts which is why I didn't include it. In fact I have the reload in my own instant of this and if Scott says it's OK I'll add it back here. But I think I asked about this on one of my other PRs and haven't received an answer yet.

mjcumming:
Yes, I was seeing an error at that line. Items were not being populated with Astro after editing the .items file and saving. I thought they only populated when the channel was updated? 

Not sure about the reload - it doesn't really cause any issues (that I know of). That being said, I am not sure I like the idea of the configuration.py file anyway. I don't think its a great solution. Ideally in 3.0 there would be a way to have scripts be able to generate their own UI's for configuration.

rkoshak:
A configuration UI was part of Kai's original vision. I'm not sure if we will get all the way there though. Time will tell. 

As for Astro, I remember seeing Astro log statements when I reloaded a file. I thought it was when I reloaded a .items file that had Items linked to Astro Channels, but now I'm wondering if I mainly saw that when I reloaded a .rules file with Astro Rule triggers. I just tried reloading my .items file and Astro did indeed not log anything. I use restoreOnStartup so I don't lose the value but I would hate to make that a requirement. I may need to do something more drastic because if you reload the .items file and Astro doesn't repopulate it until that midnight, it would break the state machine.

mjcumming:
I am thinking they do not populate - I do not persist my Astro items (maybe I should). I wonder if we can force Astro to generate channel events from a script?

volfan6415:
So if i am going to try and use this what changes do i need to make?  I see mention of the reload of a configuration and persisting astro items?

rkoshak:
I would say to just install and configure it using the docstrings in the code as a guide. If it fails to work, we can help you add the suggestions mjcumming recommends making. Or I'll set up an alternative repo until this PR get's reviewed and merged or rejected.

volfan6415:
Ok so i placed the ephem_tod.py under lib\python\community and edited the configuration.py under lib\python.

I also created the items per the explanations in the files.  

I don't see anything in my logs? And there is no Time of Day with Ephemeris rule created in my Rules in PaperUI.

Forgive me if i have done something wrong.  All of my other experience with JSR223 rules has been with javasript.  I don't even know what i should be looking for in the logs to see what the error was with rule creation.

rkoshak:
Since this is going to mostly be a debugging exercise, let's move to the OH forum. Post the Items you've created, your configuration.py, and if you created one, your Ephemeris config.

You might need to post your current version of the file because line 172 on the checked in version is blank. If I assume it's line 173 that is the error, I am going to guess there is something wrong with the Ephemeris config. There might be something error checking I can add to detect and report that if that's the case.

volfan6415:
https://community.openhab.org/t/ephemeris-time-of-day-troubleshooting/92186

Community/Ephemeris Time of Day/automation/jsr223/python/community/ephem_tod/ephem_tod.py
```diff
@@ -0,0 +1,436 @@
...
+    member of tod_group_ephem changes. When the Rule triggers, it creates timers
+    to go off at the indicated times.
+
+    For example, they full set of Groups and Items for a weekend, weekday and
+    standard bank holiday system might be something like the following:
+
+        Group Ephem_TimeOfDay_StartTimes
+        Group Ephem_Weekday (Ephem_TimeOfDay_StartTimes) { ToD="day"[type="weekday"] }
+        Group Ephem_Weekend (Ephem_TimeOfDay_StartTimes) { ToD="day"[type="weekend"] }
+        Group Ephem_Holiday (Ephem_TimeOfDay_StartTimes) { ToD="day"[type="holiday] }
```
berland, line 377: Missing " after holiday.

Community/Ephemeris Time of Day/automation/jsr223/python/community/ephem_tod/ephem_tod.py
```diff
@@ -0,0 +1,436 @@
...
+        DateTime vNight_Time "Night [%1$tH:%1$tM]"
+            <moon> (Ephem_Weekday)
+            { ToD="init"[start_time="23:00", tod_state="NIGHT"] }
+
+        DateTime vBed_Time "Bed [%1$tH:%1$tM]"
+            <bedroom_blue> (Ephem_Weekend, Ephem_Holiday)
+            { ToD="init"[start_time="00:02",tod_state="BED"] }
+
+        DateTime vBed_Time_Weekend "Holiday Bed [%1$tH:%1$tM]"
+            <bedroom_blue> (Ephem_Weekend, Ephem_Holiday)
```
berland, line 407: Missing clock configuration here.

Community/Ephemeris Time of Day/automation/jsr223/python/community/ephem_tod/ephem_tod.py
rkoshak: Further development on this PR is blocked until it gets reviewed by a maintainer. In the mean time I've re-implemented this design pattern at https://github.com/rkoshak/openhab-rules-tools. It's significantly different in configuration and use (simpler I hope).

rkoshak:

PR #284, opened 2019-11-11
Initial check in for Item Init

rkoshak:
A simple Rule that looks for Items with an init metadata and updates the Item with it's value. The intended use case is for those times when you want to boot strap an Item's state or ensure it get's initialized to a specific state rather than using a System started Rule or restoreOnStartup. Once accepted, this could be used in the Time of Day Ephemeris submission instead of having that library use it's own metadata syntax. 

It supports updating the Item only if it's different, updating the Item every time, and it can clear out the metadata once the Item is updated (a true bootstrap situation) but that only works if not using .items files.

### DIFF
--- a/Community/Item Init/automation/jsr223/python/community/item_init/item_init.py
+++ b/Community/Item Init/automation/jsr223/python/community/item_init/item_init.py
@@ -1,0 +1,60 @@
+"""
+Author: Rich Koshak
+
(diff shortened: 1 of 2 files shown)


rkoshak:
This PR is ready for review.

rkoshak:

PR #295, opened 2019-12-08
Add Selector script

CrazyIvan359:
Selector allows you to have a group of Sources that can be linked to
any of the Endpoints in the Endpoints group. This allows you to quickly
repurpose automation devices without having to edit your configuration.

I created this to simplify the repurposing of some Sonoff S31 outlets I got.
Using this I can quickly change what part of my automation they are linked to
without having to change any configuration or channel links.

~~I am submitting this as a draft because I want to have it compare accepted commands/states as the method for determining compatibility for Source -> Endpoint links. Currently only same type Items can be linked.~~
This has been addressed and is working in my tests.

@rkoshak I saw on the forum you just purchased some smart outlets, perhaps this will interest you for them. Feedback is welcome.

### DIFF
--- a/Community/Selector/automation/jsr223/python/community/selector/selector.py
+++ b/Community/Selector/automation/jsr223/python/community/selector/selector.py
@@ -1,0 +1,536 @@
+"""
+Selector allows you to have a group of Sources that can be linked to
+any of the Endpoints in the Endpoints group. This allows you to quickly
+repurpose automation devices without having to edit your configuration.
+
+- Command Sources are items that receive commands to be forwarded to Endpoints.
+- Command Endpoints are items that receive commands from Command Sources.
+- Update Sources are items that receive updates to be forwarded to Endpoints.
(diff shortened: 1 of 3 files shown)


rkoshak:
This looks like a really clever approach to solve a common problem. 

But for me, it's way less work to just redo the Channel Links or to enable/disable Rules than the configuration required to make this work for me. Perhaps if I had more than five devices I need to re-purpose it would be a different story. But even that number is misleading as I only really use one of these smart plugs during the off season, the rest live in a drawer. 

What I currently do is I have a "Tis the season" switch. When it changes, I disable/enable the humidifier Rule and Christmas lights Rule based on it's state. I then have two Items linked to the same channel. This is all I need so I won't be a good test case.

But I'm certain that there will be many users who will benefit from this. I particularly see this as being useful for other Helper Library contributors as well.

CrazyIvan359:
Thanks for the feedback Rich!

PR #311, opened 2020-04-19
Fix transformation for Javascript

J-N-K:
Signed-Off-By: Jan N. Klug <jan.n.klug@rub.de>


openhab-5iver:
There is a lot in this file that needs to be cleaned up. Since there is an [actions.js](https://github.com/openhab-scripters/openhab-helper-libraries/blob/master/Core/automation/lib/javascript/core/actions.js), the actions bits should be removed completely.

J-N-K:
And it‘s better to keep it broken because there are more things that need attention than fixing at least a part of it?

openhab-5iver:
Sorry, no... you miss my point. All references to actions need to be removed from utils.js. Fixing this in utils.js only duplicates the functionality in actions.js.

PR #327, opened 2020-06-15
Pylint changes

openhab-5iver:


besynnerlig:
I've reviewed an already merged PR. I'm not sure that you will be notified and aware about my question above @openhab-5iver

openhab-5iver:
Unfortunately, I do not see any comment or review. They do show up after merge. What was your question?

besynnerlig:
@openhab-5iver ,

> Unfortunately, I do not see any comment or review. They do show up after merge. What was your question?

In Core/automation/lib/python/core/utils.py
`LOG.warn("The 'core.utils.getItemValue' function is pending deprecation.")` and similar ...

I'm getting a lot of warnings in my log. What is the reason for the pending deprecation?

besynnerlig:
I'm not sure if you get notifications for this @openhab-5iver due to that it's closed so I post this and mention you. (Sorry if I spammed you)

openhab-5iver:
Hey @besynnerlig! No worries... I've just been busy with other things! I marked several of these functions deprecated because I'd like to remove them. I had [mentioned this when they were first included](https://github.com/openhab-scripters/openhab-helper-libraries/pull/91), but I probably should have said something to you before merging this. My reasoning is that I am preparing to recreate the helper libraries in the rule engine and this library will be the first to go. Most of these functions I think would be better documented as an example or in 'But How Do I?'. Specifically...

* kw: This is very specialized and I don't think it will be missed. Maybe something for "But How Do I"?
* iround: Instead, use int(float(str(round(test_number)))).
* getItemValue: This is a specialized function and in most cases items["My_Item"] works fine for getting the state of an Item. This would be better as an example.
* getLastUpdate: This is another specialized function that would be better as an example.
* sendCommand and postUpdate: These just eliminate the need for prepending events on these methods. People should learn to use events or we should remove it altogether.
* post_update_if_different and send_command_if_different: I think these have proven useful for migration to the rule engine by adding a boolean parameter to the actions for whether to compare to the existing value or not. Like events.sendCommand("My_Item", "ON", True) and events.sendCommand("My_Number_Item", "5.5", True, 1) (the later includes floatPrecision).

besynnerlig:
However the kw function was changed in such a way that the for loop now alters the argument hence renders wrong results. I'm not sure if I need to report that as a new issue or if it's enough to mention it here.

The problem is that the function argument name `value` is reused in the for loop.

openhab-5iver:
Oh my... fixed (#346)!

PR #353, opened 2020-08-31
fix JythonTransform

lordjaxom:
a) changed loading order of TransformationService since in (at least) OpenHAB 2.5.7 both org.openhab.core.transform.TransformationService and org.eclipse.smarthome.core.transform.TransformationService are available but the former is unused and isn't picked up by the component system

b) changed transform parameter "pathname" to "function" as in TransformationService and build "pathname" from config path + transform + function (shamelessly copied from AbstractFileTransformationService#getSourcePath)

### DIFF
--- a/Core/automation/jsr223/python/core/components/200_JythonTransform.py
+++ b/Core/automation/jsr223/python/core/components/200_JythonTransform.py
@@ -12,16 +12,19 @@ from core.log import logging, LOG_PREFIX
 TRANSFORMATION_CLASS = None
 
 try:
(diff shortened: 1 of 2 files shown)


5iver:
I had put a lot of effort into getting this working, including the changes you have, but it still did not function. It was available as an OSGi service, but did not show as a TransformationService. I've moved the script to Examples and have been working on a Java version of a Jython TransformationService. It looks like this will need to be included in the JythonScriptEngineFactory add-on, but I'm still trying to split it out for people who stay with a manual installation.

PR #172, opened 2019-06-26
Add metadata.js

openhab-5iver:
JS library for manipulating Item metadata. I used this script for testing... [metadata_test.js.txt](https://github.com/openhab-scripters/openhab-helper-libraries/files/3329427/metadata_test.js.txt). I've been using snake_case in the JS libraries... will that be a problem? My reasoning was to keep things consistent with Jython.

Signed-off-by: Scott Rushworth <openhab@5iver.com>

PR #355, opened 2020-10-14
Javascript fixed metadata functions TypeError

dreamsinmotion:
Fixing TypeError: null has no such function when calling a metadata function


5iver:
What were you doing to get this error? I've tested this extensively, but have never gotten that or any other error. [Here](https://github.com/openhab-scripters/openhab-helper-libraries/pull/172#issue-291774965) is the test script that I was using at the time the PR was merged.

PR #356, opened 2020-10-23
Enable unicode character handling without the hassle

marcinczeczko:
I'd like to propose this changes to handle unicode strings without the hassle.
It's very useful for situation where items are created via jython scripts, and item label suppose to be in the language that uses unicode characters (e.g. polish, german, etc.)

### DIFF
--- a/Core/automation/lib/python/core/items.py
+++ b/Core/automation/lib/python/core/items.py
@@ -2,6 +2,8 @@
 This module allows runtime creation and removal of items. It will also remove
 any links from an Item before it is removed.
 """
+from __future__ import unicode_literals
+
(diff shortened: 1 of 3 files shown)


5iver:
Thank you for submitting this, but there are several reasons why I will not include unicode_literals. You can easily look those up on the Internet. Unicode is very easy for users to deal with once they have an understanding of how to use it in their rules. I've also recently submitted PRs that allow unicode to be used for Item names, rule names, logging, etc., so there should not be a need to change the code. I have a documentation update coming, and I will try to fill in some more details about the use of unicode. I'll be very interested to know what you think.

PR #369, opened 2020-12-06
Remove items attributes

5iver:
Being able to get the state of an Item with this format...

```
item_state = items.my_item
```
... has caused many issues. Users have also reported that this functionality can randomly stop working.  I do my best to not make breaking changes and to always provide backwards functionality, but I don't see this adding anything that doesn't already exist, so I don't see any harm in removing it. 

Signed-off-by: Scott Rushworth <github@5iver.com>


stagf15:
@5iver, I just updated my helper libraries the other day (after this commit) and noticed several of my items/rules seemed to be broken.  When I dug deeper I realized that the ```__init__.py``` file had nothing in it.  I replaced it with the old ```__init__.py```, restared, and everything worked as it should. I use ```item_state = items.my_item``` throughout my python rules.  Is this the cause, and what should I be using?  Thanks.

5iver:
Sorry for the headache @stagf15... this is definitely the cause. Instead of items.The_Item_Name, use items["The_Item_Name"]. The __init__.py for the core package should now be empty.

stagf15:
Ok! I'm just happy to know what to fix to bring it up to the current standard.  Thank you very much for all of your hard work. 

I've been reading up on OH3 and jython incorporation, BTW... I loaded OH3 on my windows test machine assuming jython and jsr223 rules would be totally integrated, and was surprised that it wasn't, so I've put OH3 on hold until it's ready.  Any update on when you think that will be?

PR #382, opened 2021-02-22
Enable verbose testing

politas:
Change to `core.testing` to allow a choice to get more detailed testing results, using a temporary file to store test results.

Calling `core.testing.run_test()` with 'verbosity=<num>` as a parameter will now open a temporary file, redirect the TextTestRunner output to it, and then feed the test results to the log. If no verbosity is supplied, the result is unchanged.


politas:
Sorry, meant to PR to CrazyIvan's ivans-updates branch

